PR #528, opened 2015-01-22
Dev

Starwaster:
- Ascent Autopilot recomputes desiredFlightPathAngle as needed (instead
of recording lastDesiredFlightPathAngle ) in DriveCoastToApoapsis()
- Ascent Autopilot freezes all steering in DriveCoastToApoapsis() when
performing an apoapsis boost (ThrottleToRaiseApoapsis())
- Docking Autopilot gives increased priority to closing lateral distance
to docking axis when further away than 1 meter.


sarbian:
The diff shows the whole MechJebModuleAscentAutopilot.cs as changed. I guess you changed the linefeed of the file or your git core.autocrlf config.
Can you check and re submit ? I'd rather have diff that show the only actual change

Starwaster:
Yeah I'm not sure what happened there or how to get back to how it was before. Originally I was just trying to get git to stop tracking the userprefs file and that went badly because it removed/re-added every single file in the repository and so I reverted that but was left with a conflict in that file and in fixing the conflict I munged the line endings.... I'll see if I can fix. (kind of scared to. I hate Git)

sarbian:
http://www.sourcetreeapp.com/   makes most on git way more intuitive.

PR #538, opened 2015-01-29
Landing autopilot modularization, each step is implemented in a separate class

Meumeu:
This will simplify future changes, also small course corrections are performed on RCS


sarbian:
I took more time to look at the change today and I love the idea :)

chriscamacho:
Just think, if only all these modules could be scripted....

PR #548, opened 2015-03-21
Reduce transient error in attitude controller

xytovl:
Attitude controller now uses feed forward to better follow a moving
target.

Testing with a basic rocket shows steering error reduced to 4° maximum at start of gravity turn, compared to 9° with previous version.


sarbian:
Works fine in my tests too. Thanks !

Issue #550, opened 2015-03-23
New atitude control and kill rot


futrtrubl:
The new system doesn't seem to be able to handle Kill Rot. All 3 axes go full on but the vessel's spin isn't stopped, it may even be made worse. Other modes where it has a definite position to face work perfectly on the same craft.

sarbian:
Which dev version did you test with ? There was a fix that may be related to that in #413 a few hours after the initial fix.
@xytovl  ?

lamont-granquist:
Closing this as an old MechJeb issue.  Many MechJeb issues have had little interest in years, have been fixed for years, do not include adequate replication steps, refer to old problems which are no longer applicable, or are difficult to determine what the problem is.  This issue is being closed for one of those reasons.  We apologize for any inconvenience, but keeping the TODO list tidy helps the developers.

If this bug/issue is still a problem, please open a new issue.  For bugs please try to include a [Minimal, Complete, Verifiable Example](https://stackoverflow.com/help/mcve) that explains all the steps required to replicate the issue.  A link to the KSP.log file should be ideally included, but is often not sufficient information.  Screenshots or short videos are often the best way to show a bug.

PR #551, opened 2015-03-23
Disable attitude controller feed forward in kill rot mode

xytovl:
If we get other situations where it causes problems we might have to make it opt-in, as a parameter in attitudeTo.
Fixes issue #550


futrtrubl:
I don't think it will be a problem for other situations as kill rot is the only place it doesn't have an explicit desired attitude. But I will test others as I am able like translatron.

PR #574, opened 2015-05-02
Remove 1.0.2 warning and change pressure unit to kPa

AndreyATGB:
getStaticPressure seems to return kPa now, I just changed the unit
display to kPa. It would be possible to divide the result by 101.325 and
get back atm, though I'm not sure how it affects FAR.


sarbian:
I wrote all those change days ago on my local copy but I need to fix the science unlocking and dv stats not working for some before I push it.

PR #592, opened 2015-05-20
Lcg/low altitude landing safety

lamont-granquist:
if we press one of the landing buttons when we're below the deceleration
burn altitude bad things can happen.

* if we're off course by > 150m then we get stuck in CourseCorrection and
  will reliably lithobrake
* i've seen craft get stuck in "coasting towards deceleration burn" and
  lithobrake but replication is unreliable

to address both cases, add shortcuts so that we wind up in the logic
in DecelerationBurn that selects KillHorizontalVelocity or FinalDescent
correctly.

also I've had two craft lithobrake at 50x warp, which is not reliably
reproducable and happens too fast to see what mode its in.  it may be
due to the 10-seconds-to-impact extrapolation not taking into account
gravitational acceleration on very low initial velocities and low
heights.

### DIFF
--- a/MechJeb2/LandingAutopilot/CoastToDeceleration.cs
+++ b/MechJeb2/LandingAutopilot/CoastToDeceleration.cs
@@ -77,14 +77,13 @@ namespace MuMech
                     }
                 }
 
-                /*
-                // Sometimes (on bodies with a thick atmosphere) there is no need for a decleration burn. Check for this so that it is possible to transition into the final decent step.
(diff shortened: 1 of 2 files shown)


lamont-granquist:
Ref: http://forum.kerbalspaceprogram.com/threads/12384-PART-1-0-2-Anatid-Robotics-MuMech-MechJeb-Autopilot-v2-5-1/page1028

sarbian:
Looks fine. Many thanks :)

PR #614, opened 2015-06-06
Added GameData to packaging path.

mathuin:
The .zip and .tar.gz for MechJeb2 currently extract to ./MechJeb2
which means they cannot be extracted from the KSP root directory like
most mods.  This change adds the ./GameData directory to the head of
the path, which means the mod can now be installed from the root
directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,18 +64,18 @@ build/%.dll: ${MECHJEBFILES}
 		-resource:build/Resources.resources,MuMech.Properties.Resources.resources
 
 package: build ${MECHJEBFILES}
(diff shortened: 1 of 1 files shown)


sarbian:
I can merge it but it won't change a thing for the jenkins generated zip. 

Changing that requires work on 2 different build system (MJ Jenkins, mine) and the CKAN build. I am not really motivated to do all that.

mathuin:
Are the configuration files for Jenkins and CKAN accessible to me?  I would be willing to change them to match this change.

PR #624, opened 2015-06-13
Fix differential throttle

Meumeu:


sarbian:
oops :confused:
Thanks !

PR #657, opened 2015-08-15
Dev

Raf04:
1- Improvement of PID to provide a limitation of angular velocity. This does the spins  slower, but best controlled. Also it reduces fuel usage by RCS.

2- Fix error in "VesselState".  AttitudeController thinks it can get some torque from not ignited engines

there has been a mistake, I wanted to merge from last Dev...


sarbian:
Thanks for the patch but wrong target, you choose master instead of dev

sarbian:
And I don't see your patch on your fork. Strange :(
If you have a problem with the marging send me a zip of the patched code and I ll handle it

Raf04:
Your Zip...

https://www.dropbox.com/s/zoa4zah1sw36tyt/AVlimiter.rar?dl=0

Raf04:
I tested version # 505 in ROV. and continues adding torque from not ignited engines.
Any of following options seem to work well:
                if (!e.getFlameoutState)
                if (!e.flameout)
                if (e.isOperational)
                if (e.getIgnitionState)

sarbian:
e.isOperational is the one I used.

Raf04:
Sorry for my English, I mean that any of the options seem do not work well

PR #659, opened 2015-08-20
Add InfoItems for periapsis in target SoI and minimum DV required for capture by target.

Meumeu:
So we don't have to check the map view to see the periapsis when doing an interplanetary transfer.

Also ArchLinux uses mcs instead of gmcs, the compiler can be overridden by an environment variable.


sarbian:
Nice. quite useful :+1:

PR #667, opened 2015-09-08
Added 'toggle ascent guidance' window item button and made 'Extend all solar panels' a toggle button

StevilKnevil:
To make it easier to have a custom window with some of the functionality of the Ascent Guidance, I've expose the toggle ascent guidance as a button that can be exposed on custom windows.

Note that to simplify and reduce duplicated code, the information label has been modified, though this has the bonus side effect of keeping the positioning of the UI within in the ascent guidance window stable:

image

On this fork there is also a change that makes merges the 'Extend/Retract all solar panels' button a single toggle button rather two seperate buttons.

### DIFF
--- a/MechJeb2/MechJebModuleAscentGuidance.cs
+++ b/MechJeb2/MechJebModuleAscentGuidance.cs
@@ -61,6 +61,26 @@ namespace MuMech
             }
         }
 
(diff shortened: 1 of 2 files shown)


sarbian:
Looks fine. Thanks :)

PR #668, opened 2015-09-09
Fix for NavBall target not updating when AscentGuidance window is hidden

StevilKnevil:
Moved the NavBall logic into the AscentAutoPilot so that it can be updated regardless of whether the UI is present. Also means that the AutoPilot can be 'Enabled' but not 'Engaged':

Enabling just means it will update (e.g. the nav ball, pluss optionally the vessel control)
Engaged means it will actually control the vessel

One minor issue is that if you have the ascent path shown on the navball and then right-click disable the MechJeb module on the vessel, the ascent target remains on the navball (though is no longer updated). I haven't been able to figure out the correct hook for when than happens (i.e to remove the autopilot user that I add in AscentGuidance.OnStart())

### DIFF
--- a/MechJeb2/MechJebModuleAscentAutopilot.cs
+++ b/MechJeb2/MechJebModuleAscentAutopilot.cs
@@ -82,32 +82,75 @@ namespace MuMech
         AscentMode mode;
         bool placedCircularizeNode = false;
         private double lastTMinus = 999;
+        private bool engaged = false;
 
-
-        public override void OnModuleEnabled()
-        {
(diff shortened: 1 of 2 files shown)


sarbian:
FYI your OnDestroy was not called because you used virtual instead of override in your declaration.
I fixed it there https://github.com/MuMech/MechJeb2/commit/934fcc9e6906e93d0e2232ec06c6485e5cb00b00

StevilKnevil:
Bah - of course! Thanks for fixing that up :)

PR #674, opened 2015-10-11
MechJebModuleStagingController: improve autostaging when there is uneven resource consumption

billkerbinsky:
This module looks at both engine state and resource presence to determine
whether to inhibit staging.  In the absence of crossfeed, engine state
should be sufficient, but for engineless drop tanks (and hybrid configurations
like the never-flown Saturn V 4/260 where there's a crossfed drop tank
on top of a solid-fuel booster) we should be able to drain those tanks, too.

However, looking at the resources required by all engines means that
if there's a slight propellant imbalance we won't stage if we're just
out of oxidizer; to fix this, exclude engines which would be dropped
in the next staging from the resource evaluation.

### DIFF
--- a/MechJeb2/MechJebModuleStagingController.cs
+++ b/MechJeb2/MechJebModuleStagingController.cs
@@ -130,9 +130,11 @@ namespace MuMech
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


sarbian:
Let's see how it fares. 
Thanks for the PR :)

PR #680, opened 2015-10-25
MechJebModuleThrustController: Add electric throttle limiter

billkerbinsky:
Craft powered by electric engines can become uncontrollable when their
probe core is starved of power.  When in this state, all available
power is immediately consumed by the engines and there is no way to
adjust the throttle.  This can result in much longer burns than
desired.

This change adds an optional proportional throttle limiter which will
scale back the throttle when the vessel is low on ElectricCharge.  It
can safely be left enabled at all times, as it is only active when an
active engine uses ElectricCharge as a propellant.

There are two parameters: ElectricLimitLo and ElectricLimitHi,
expressed in the UI as percentages of total ElectricCharge capacity.

While in operation, the limiter can be in one of three states,
indicated by the color of the limiter's label:

 - Red: If the vessel's ElectricCharge level is below ElectricLimitLo
   (default 5%), the throttle is limited to zero.

 - Yellow: If ElectricCharge is between the two limits, set the
   throttle in proportion to where it sits within that range.

 - Green: If the vessel's ElectricCharge level is above ElectricLimitHi
   (default 15%), the throttle is unlimited; naturally, other limiters
   may still reduce thrust below 100%

While operating in the yellow state, the limiter finds a throttle
setting which consumes only as much power as the vessel can generate.

Tuning:

You need to set the threshholds high enough that a sudden increase in
electrical demand won't consume all power before the throttle goes to zero.

The low limit is most critical; for solar-powered craft, you should
set it high enough that batteries won't run down to zero during the
longest expected orbital eclipse period.

Setting the "Hi" limit is more a matter of taste; lower levels mean
you can make longer sprints at full power before the limiter kicks in,
but also increase the risk of running out of power entirely.

### DIFF
--- a/MechJeb2/MechJebModuleAscentGuidance.cs
+++ b/MechJeb2/MechJebModuleAscentGuidance.cs
@@ -109,6 +109,7 @@ namespace MuMech
             core.thrust.LimitAccelerationInfoItem();
             core.thrust.LimitThrottleInfoItem();
             core.thrust.LimiterMinThrottleInfoItem();
+            core.thrust.LimitElectricInfoItem();
             GUILayout.BeginHorizontal();
             autopilot.forceRoll = GUILayout.Toggle(autopilot.forceRoll, "Force Roll");
             if (autopilot.forceRoll)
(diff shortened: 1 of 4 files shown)


sarbian:
I am a bit concerned by the fact that it check the whole ship each frame for electric charge when running but other than that it look fine. I know other part of the code does it but it slow down things more and  more.
Thanks for your work !

PR #698, opened 2016-01-24
Missing *.cs files in Makefile

Meumeu:


sarbian:
Wouldn't it be easier to simply do 
MECHJEBFILES := $(wildcard MechJeb2/**/*.cs)    
?

Meumeu:
wildcard does not work recursively but
```
MECHJEBFILES := $(shell find MechJeb2 -name "*.cs")
```
works, at least under Linux.

sarbian:
Two wildcard should do the lookup in the subdirs. Look at my post again, the first time github formatting removed some char

PR #730, opened 2016-04-29
fix wandering camera when mouse over MJ menu

Starwaster:
* fix wandering camera when mouse over MJ menu. (change locked ControlTypes to ControlTypes.ALLBUTCAMERA)


Starwaster:
Looks like there WAS a way after all.

sarbian:
Did I said there was none?

Starwaster:
No, *I* did. (sort of)

PR #770, opened 2016-07-04
add button to use stock MOI

lamont-granquist:
this also converts the angular momentum computation to use the MOI

it does not use the full inertia tensor so the angular momentum may
be slightly less accurate, but the diagonal elements should
dominate?

### DIFF
--- a/MechJeb2/MechJebModuleAttitudeAdjustment.cs
+++ b/MechJeb2/MechJebModuleAttitudeAdjustment.cs
@@ -178,8 +178,8 @@ namespace MuMech
                     GUILayout.EndHorizontal();
 
                     GUILayout.BeginHorizontal();
-                    GUILayout.Label("MOI", GUILayout.ExpandWidth(true));
-                    GUILayout.Label("|" + vesselState.MoI.magnitude.ToString("F3") + "| " + MuUtils.PrettyPrint(vesselState.MoI), GUILayout.ExpandWidth(false));
+                    GUILayout.Label("MOI calc", GUILayout.ExpandWidth(true));
+                    GUILayout.Label("|" + vesselState.calcMoI.magnitude.ToString("F3") + "| " + MuUtils.PrettyPrint(vesselState.calcMoI), GUILayout.ExpandWidth(false));
                     GUILayout.EndHorizontal();
 
                     GUILayout.BeginHorizontal();
@@ -200,6 +200,7 @@ namespace MuMech
             }
 
             MechJebModuleAttitudeController.useCoMVelocity = GUILayout.Toggle(MechJebModuleAttitudeController.useCoMVelocity, "Use CoM velocity instead of stock");
+            vesselState.useStockMoI = GUILayout.Toggle(vesselState.useStockMoI, "Use stock MoI");
 
             MechJebModuleDebugArrows arrows = core.GetComputerModule<MechJebModuleDebugArrows>();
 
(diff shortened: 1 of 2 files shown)


lamont-granquist:
`vesselState.MoI` is still what all the rest of the code should use.  this builds a button to change the meaning of that to be either MJ's calc or the stock calc.  in order to try to eventually eliminate MJs MOI calc i also made it use the MOI to compute the angular momentum.  that may or may not work.  but.  this is not enabled by default, and the point is to let us try pressing it and see how good/bad it gets.  right now there's issues with stock MOI so this button shouldn't be used (yet).

sarbian:
Ok, merged. 

What is the problem with the stock MOI ? Because if the problem and fix are known we could have MJ use the fixed version while we wait for the next KSP version.

lamont-granquist:
Problem isn't fixed yet.  I think there's an incorrect rotation being applied to the part inertia tensor before its being added to the rest of the parallel-axis terms.

An essentially single-part big tank example:

http://i.imgur.com/IAXux7h.jpg

But NK is busy with Unity 5.4 and 1.2.x stuffs right now...

I'd actually consider going the other way and just copying MJs code into KSP...  When I wrote the version that NK picked up and put in core I think I made some bad assumptions about the code quality of Unity's Vector and Matrix libraries...

Although if we got the core code into MJ for awhile we could swap between the two and benchmark the results, but I don't think we need to actually commit code to MJ to do that, that could just be done on a dev branch somewhere...

sarbian:
I did some test with a fixed version of the stock code and the result are ... mixed. The value are totally different from MJ's value ( https://i.imgur.com/3cMlQkI.png ) but the handling is not improved in all cases (that may comes from other problems).

So at this point I think I will do some "real" experiments. Build an heavy ship with only a reaction wheel, make it turn on 1 axis for 20s, check the angular vel and find the MoI from that. That way I can check if any of the three MoI version is the right one.

lamont-granquist:
Well that's curious.  It looks like the part inertiaTensor contribution is low.  I don't know why it would be off by a factor of two, the rest of the calculation just requires part mass and radius-from-CoM calculations to be correct.

And yeah, I think it would be good to check MOI and Torque calculations against what the Physics should be (and against what the Unity physics engine does).

lamont-granquist:
So there's something wrong with the Torque calculation from RCS.   I've got a ship which has 4 x 69/111N class RCS thrusters on this TLI stage of mine.  They're supposed to be .084kN thrusters (MMH+NTO).  They're offset back from the CoM by 1.2m and they're at a radius of 0.725m.

For roll all 4 of them will fire at a distance of 0.725m from the roll axis for:   4 x 0.084 x 0.725 = 0.243 kNm
For pitch/yaw 2 will fire at a distance of 1.2m from the pitch/yaw axis for:  2 x 0.084 x 1.2 = 0.200 kNm

What I'm seeing in the display in MJ is [ 0.811, 0.773, 0.811 ] instead of the expected [ 0.200, 0.243, 0.200 ].

I tried removing the "movingCoM" calculation and just using CoM and that didn't change anything.

I wonder if somehow its adding up all 4 of the jets on these quad blocks?

Anyway, too late to add debugging code tonight to see if I can determine what is going on....

sarbian:
As I said on IRC the test shows that the stock MOI code + fix gives the proper value. However I need to talk to @NathanKell before pushing that code.

I just pushed some code to fix an RCS problem (28af35f144f56672cc08d46f821db6f2b72736e8) but I doubt it will fix your problem.
Edit/Add a custom window (Custom Window Editor if you don't know) and then from the "Vessel" category add the "Troque Compare" item. It will show you something like this picture. The first column displays the RCS torque as computed by the "old" MJ code and the second will show the RCS torque given by the stock code. For most ship they are the same but for some they differ a lot. I did not find a pattern yet but this is the the most simple ship that I have that shows it :
https://i.imgur.com/hAfEZQi.png

PR #772, opened 2016-07-10
Do not throttle with unstable ullage

lamont-granquist:


lamont-granquist:
wow that was quick =)

MechJeb2/VesselState.cs
```diff
@@ -1133,14 +1181,36 @@ public void Update(Vector3d c, Vessel vessel)
...
                 float alt1 = (float)(vessel.altitude + TimeWarp.fixedDeltaTime * vessel.verticalSpeed);
                 atmP1 = (float)(FlightGlobals.getStaticPressure(alt1) * PhysicsGlobals.KpaToAtmospheres);
             }
 
+            public void CheckUllageStatus(ModuleEngines e)
+            {
+                if ((!e.EngineIgnited) || (!e.isEnabled))
+                {
+                    return;
+                }
```
lamont-granquist: There is a theoretically possible edge condition I didn't bother with -- if you could have an engine which is throttled up transition to a unstable ullage situation, then MJ would kill the throttle on an already throttled up engine.

I believe we can safely ignore that because engines should always have enough power to ullage themselves (and if they don't that's a horribly designed rocket -- incredibly small engines and incredibly large mass? -- and you kinda deserve what you get....)

PR #773, opened 2016-07-10
Handle unstable throttled engine edge condition

lamont-granquist:
Decided to try to address this edge condition even if it seems very
unlikely.

### DIFF
--- a/MechJeb2/VesselState.cs
+++ b/MechJeb2/VesselState.cs
@@ -1192,7 +1192,9 @@ namespace MuMech
 
             public void CheckUllageStatus(ModuleEngines e)
             {
-                if ((!e.EngineIgnited) || (!e.isEnabled))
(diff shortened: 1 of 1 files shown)


lamont-granquist:
And the edge case I'm mostly concerned with is the Player manually hitting the throttle on an engine which is not "Very Stable".  If the autothrottle stuff is running then, we'd cut the throttle back to zero and make certain to kill the engine because RF only might have killed the engine.

I believe with this code if the user manually throttles up in an unstable state with autothrottle routines running we should roll with it, and hopefully it all turns out well.

This might be important because I only considered "Very Stable" to be stable for the purposes of MJ, because its a robot and it should have high standards...

I also used requestedThrottle rather than thrust or fuel flow or something like that because once the engine is requested to throttle up I suspect its too late to try to cut it out for the user?

sarbian:
Yes, it should be too late to do anything. 
Thanks for the patch!

PR #774, opened 2016-07-11
Add auto-ullaging with RCS

lamont-granquist:
If RCS is on and we're in a condition where we're suppressing the
throttle due to ullage, fire the RCS jets in order to ullage.


lamont-granquist:
Okay, I think that's it for today.

This one comes with a gifv:

https://gfycat.com/EcstaticColorfulKinglet

PR #775, opened 2016-07-11
Prevent Unstable Ignition Menu Button

lamont-granquist:
Improves the user-experience for preventing unstable ignitions and doing RCS auto-ullage.

Adds a button to the Utilities menu -- by default only when RealFuels is installed so as not to confuse stock players -- that is on by default.

When MJ is suppressing the throttle due to ullage concerns it will light up green just like "Prevent jet flameout" does.

Also has a fix so that RCS auto-ullaging works with manual throttle.

### DIFF
--- a/MechJeb2/MechJebModuleThrustController.cs
+++ b/MechJeb2/MechJebModuleThrustController.cs
@@ -441,6 +441,13 @@ namespace MuMech
             // RealFuels ullage integration.  Stock always has stableUllage.
             if (!vesselState.stableUllage)
             {
+                if ( targetThrottle > 0.0F && throttleLimit > 0.0F ) {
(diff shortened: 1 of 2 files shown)


lamont-granquist:
Node Execution:

https://gfycat.com/DizzyLoneBarnowl

Manual Throttle w/RCS auto-ullage:

https://gfycat.com/MadeupClearHawaiianmonkseal

Manual Throttle w/o RCS enabled (just shows that the menu goes green to show the user why the throttle isn't responding):

https://gfycat.com/RingedBowedLaughingthrush

lamont-granquist:
User on IRC mentioned "stuck throttle" on build 619 so I figured they needed some indication that MJ was blocking throttle up.

This all feels like fairly slick UX to me, so I think I'm done now...

PR #777, opened 2016-07-19
Better launch inclination tracking

lamont-granquist:
essentially teaches MJ about great circle ground tracks

what the old code did was to take the launch latitude and save it
and then build the heading off of it.  so that if you launched
at 80 degrees (10 degrees inclination / 10 degrees N) then MJ would
continue to burn that direction for the entire burn.

it kinda sorta got burns kinda sorta right for kerbin launching from
the equator with really fast high TWR burns that raised apoapsis quickly, but
for anything approaching real burns the old algorithm was terrible.

with the new algorithm i had to sacrifice pathological launches in
favor of non-pathological launches.  once you've built up more than
200 m/s of surface velocity (taking into account rotation of the body)
then MJ won't let you direct a south-going rocket to the north any
more (and vice versa).  MJ has to use some kind of heuristic to
determine when it thinks it has passed the apex of a great circle
ground track and has to start going the "negative" direction from what
was typed into its inclination box.

### DIFF
--- a/MechJeb2/MechJebModuleAscentAutopilot.cs
+++ b/MechJeb2/MechJebModuleAscentAutopilot.cs
@@ -205,7 +205,7 @@ namespace MuMech
             //during the vertical ascent we just thrust straight up at max throttle
             if (forceRoll)
             { // pre-align roll unless correctiveSteering is active as it would just interfere with that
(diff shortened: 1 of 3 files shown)


lamont-granquist:
launch_inclination

lamont-granquist:
Note that I'm reasonably certain that someone, somewhere will have some kind of issue with this and that there will be someone showing up in github issues or on the forum complaining about it.

Without them actually showing up with their problem, though, I don't know what should be fixed, or if they just need to not try to do incredibly silly things...

sarbian:
As long as it works within reasonable expectation I am fine with it. MJ users will always find way to break things anyway :)

lamont-granquist:
I figured you knew your own userbase =)

PR #778, opened 2016-07-20
Handle ullage ArgumentException error

lamont-granquist:


lamont-granquist:
I'm not super duper happy with catching the exception but it should turn itself off after catching one of them.

Reflection in C# is odd and I don't really understand why it doesn't happen back when we try to grab the field and waits until we try to get the value...

sarbian:
It does not makes sense for me either. I feel that something else (version ?) is going on.

lamont-granquist:
I did find a SO that suggests this kind of tactic is appropriate for SetValue:  http://stackoverflow.com/questions/34433043/check-whether-propertyinfo-setvalue-will-throw-an-argumentexception

The alternative would be to go poking around in RF more to determine if ullage was disabled or not, but it feels like the less coupling to RF APIs, the less likely it'll break...

PR #779, opened 2016-07-23
Use ship altitude instead of final alt for ascent heading

lamont-granquist:
Now that we're continuously updating the heading based on the vessel
position, it makes more sense to use the vessel's altitude as the guess
for our direction rather than the final orbital velocity.

This trades some wildly inaccurate numbers for some slightly inaccurate
numbers.

If you want to ascend directly to a large apoapsis with a large
inclination then MJ will try to shoot for the horizontal velocity of
the apoapsis while its still on the ground.  At the absurd limit if
you punch in the SOI limit of the body you're lifting off of you'll
be asking to putter along at a few m/s and that is definitely
insufficient for a launch velocity.

This algorithm gets it a little bit wrong, but it will self correct as the
rocket ascends and it continually updates.

As a refinement we could go back to injecting the target altitude again
and then calculating the velocity at the periapsis of an orbit with the
periapsis at the surface and the apoapsis at the desired height, but i'm
not sure we need to and i haven't quite convinced myself that would be
any more correct.

### DIFF
--- a/MechJeb2/MechJebModuleAscentAutopilot.cs
+++ b/MechJeb2/MechJebModuleAscentAutopilot.cs
@@ -205,7 +205,7 @@ namespace MuMech
             //during the vertical ascent we just thrust straight up at max throttle
             if (forceRoll)
             { // pre-align roll unless correctiveSteering is active as it would just interfere with that
-                double desiredHeading = OrbitalManeuverCalculator.HeadingForLaunchInclination(vessel, vesselState, desiredInclination, desiredOrbitAltitude);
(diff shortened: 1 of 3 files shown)


lamont-granquist:
there's a bigger issue with kerbin-launched rockets at e.g. 90 degrees winding up at 88.6 degrees instead of 90 degrees in two-burn ascents on kerbin because we really want to be targeting that we hit a launch inclination of 90 degrees at the end of the first burn, not at the end of the circularization burn (and trying to adjust the heading at the circularization burn may be incredibly bad if the circ burn is over the north pole -- so we really need to be more aggressive about turning right at launch when we're at the lowest latitude we'll be at....)

lamont-granquist:
(but none of these commits have made that problem any worse than it was previously.....)

PR #781, opened 2016-07-24
launch inclination fixes for Kerbin

lamont-granquist:
this patch is a bit of hackery to make two-stage burns work correctly
on Kerbin, also tested on Mun and even Gilly.

on Earth we don't need this so much because we tend to do single burns
and the Earth is really big.

the fracDV calculation essentially selects .85 as the frac to apply
to Kerbin while fading out to a correction of .00 on Earth.

the correction factor also fades out as we launch and the deltaHV
value trends down.  that means we correct stronger sooner which
should(?) reduce the steering losses from high inclination launches.

as the deltaHV value whips around we stop doing this calculation as
well (which is fairly important because deltaHV can get large again but
point in the wrong direction under some conditions).

the 0.85 value is frighteningly large, but i had to do that to get
rockets to launch correctly on Kerbin.  i can't really explain it with
numberical arguments but the rockets launch well.  on Kerbin with 175m/s
of coriolis velocity at KSC it only costed that plus an extra 16m/s of
steering losses to hit a 90.01 inclination orbit compared to a 0 degree
one.

tested with 90 degrees and 135 degrees inclination on Kerbin with SLTs
ranging from 1.1 to over 2.5.  on Gilly with a 2000 launch TWR it didn't
work because that's just insane -- one puff and you launch whichever
direction you were pointed on the surface -- but after clamping the
throttle to 1% it worked fine.

since we track continuously as we launch now any inaccuracy in the
initial guess will get corrected over time.

rockets with truly insanely high TWRs and super-short takeoff burns to
raise the apoapsis will still not work well.  the two possible
strategies there will be to simply throttle down -- or else i think we
could fade this algorithm out as the TWR gets stupidly big.  not sure
how much it matters though since at extreme TWR attitude control doesn't
help you any and you launch like a gun based on whatever your attitude
was due to the surface contour.

### DIFF
--- a/MechJeb2/OrbitalManeuverCalculator.cs
+++ b/MechJeb2/OrbitalManeuverCalculator.cs
@@ -245,29 +245,30 @@ namespace MuMech
             // fall back to tracking desiredHorizontalVelocity
             if ( Vector3d.Dot(desiredHorizontalVelocity.normalized, deltaHorizontalVelocity.normalized) < 0.90 )
             {
+                // it is important that we do NOT do the fracReserveDV math here, we want to ignore the deltaHV entirely at ths point
                 return MuUtils.ClampDegrees360(180 / Math.PI * Math.Atan2(Vector3d.Dot(desiredHorizontalVelocity, east), Vector3d.Dot(desiredHorizontalVelocity, north)));
-            } else {
-              // when doing two-burn ascents we should really integrate the first burn up to the target altitude, and then
-              // we should target hitting the desired inclination angle at that point and work backwards.  Instead we have
-              // this somewhat janky algorithm that I dreamed up.  We lop off a good chunk of the desired horizontal velocity
-              // so that the angular change caused by the deltaHV correction makes the angle larger.  The 8400 magic number
-              // here shuts down this correction for launches from Earth (and from Eve?).  The 0.85 number clamps the fraction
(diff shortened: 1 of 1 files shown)


lamont-granquist:
left the "numberical arguments" typo in there because it seemed appropriate...

lamont-granquist:
actually one second...

lamont-granquist:
okay, slightly more readable to me now

lamont-granquist:
A couple more comments for posterity if anyone is debugging this months from now:

Originally I thought this would be as simple as adjusting the desiredHV to subtract off the circularization burn.  I was expecting a fracDV value somewhere around 0.5-0.3 rather than 0.85, but values that low led to lots of inaccuracy in polar orbit inclinations.  Some of the increase in that value might be due to the "fadeout" that I implemented so that towards the end of the burn it stops correcting so hard, and corrects more at the start.  It also occurs to me now that it might be better to consider scaling the deltaHV up (in other words, artificially increasing the coriolis velocity at the start of the launch to artificially apply a launch angle correction) and it might be useful to clamp the maximum angle that this correction can apply (although it applies about 30 degrees of correction on Kerbin with a 90 degree launch so we'd need to allow at least that much).  Integrating a full simulation of a launch and numerically finding precise answers to these kinds of questions would obviously be best, but that would take a ton of work.

PR #788, opened 2016-09-03
Script module - Version 0.2

SPD13:
This is an update to the Script Module
- New action to start action groups from scripts
- New action to wait for an event (only altitude for the moment)
- New action to Load (and start) a script in another slot.
- New action to manipulate the node tolerance
- Landing action can now use target landing
- A lot of work on the Quicksave, to automatically restart where the script was paused after the quicksave is loaded
- A lot of work on the Vessel switch, when switching vessels, the mechjeb core will save the active script on a temp config and load it again on the new vessel and continue the script.
- Fix on all the modules to use the "OnUpdate" action rather than the WindowGUI to determine if the action should be terminated


sarbian:
I ll let you post in the thread since I am away for a few days and only have a phone

SPD13:
Hi,
Sure, no problem, i'm preparing a video to demo the new features.
Thank you

Best

2016-09-03 13:25 GMT+02:00 sarbian <notifications@github.com>:

PR #809, opened 2016-10-28
Docking tweaks

Starwaster:
* Tweaked 'wrong side' detection. If vessel docking port zSep is behind target port, use half our own bounding box size. If we're past that point then go to wrong side backing up.  Otherwise just back straight away from target port
* Reverted FixSpeed() changes


MechJeb2/MechJebModuleDockingAutopilot.cs
```diff
@@ -318,10 +317,15 @@ void InitDocking()
 
 			if (zSep < 0)  //we're behind the target
 			{
-				if (Math.Abs (zSep) > safeDistance * 0.5f)
+				// If we're more than half our own bounding box size behind the target port then use wrong side behavior
+				// Still needs improvement. The reason for these changes is that to prevent wrong side behavior when our
+				// port slipped behind the target by a fractional amount. The result is that rather than avoiding the 
+				// target ship we end up trying to pass right through it.
+				// What's really needed here is code that compares bounding box positions to determine if we just try to back up or change sides completely.
+				if (Math.Abs (zSep) > vesselBoundingBox.size.magnitude * 0.5f)
```
Starwaster: The issue here, for the curious, is that if the vessel port gets behind the target port then the Docking AP has been assuming that the entire vessel is on the wrong side of the target vessel and sometimes tries to ram right through it. Sometimes all that's needed is to back us away a little bit or back to safe distance.

This change makes it check zSep against half of the size of the vessel's bounding box. This works better but what's really needed is to compare the position of the two bounding boxes and the two ports involved.

PR #813, opened 2016-11-06
Add new Shaders for Module DebugArrows on win64

Nabla-r:
We can't create Shaders from binary data anymore (thanks unity...), so I
created a Unity project (MechJeb Shaders.zip) and imported both shaders,
then built a bunch of AssetBundles (Right click in Project view, last menu
entry) and copied the win64 variant next to the dll. ModuleDebugArrows then
reads the bundle and loads the shaders.

Sooo, the problem is, I could'nt bring Unity to compile a shader for all
three platforms! This means: The win platforms get *only* directx subprograms,
the osx's opengl and linux opengl and metal(?).

I see two solutions:
1) Find out how to manipulate the shader and assetbundle building in
Unity, maybe with more editor scripts
2) Load all three bundles at runtime and check, what shader is supported
and use that.

What to do? Hmmmm, is there a third, maybe simpler, one?


sarbian:
Thanks for the code but I can't merge it as is. You call Unity method from constructor and that means that they can be called outside the main thread. That s a bad idea in Unity and can break the debugger.
As for the shader it s a matter of settings in the Unity preference. You need to disable auto API in the player settings and set the proper API.

Also I have some code for that already here :(

Nabla-r:
> Thanks for the code but I can't merge it as is. You call Unity method from constructor and that means that they can be called outside the main thread. That s a bad idea in Unity and can break the debugger.

Ah, I see, this explains the crashes I had!

> As for the shader it s a matter of settings in the Unity preference. You need to disable auto API in the player settings and set the proper API.

Thank you, found 'em.
 
> Also I have some code for that already here :(

No worries, it was fun experimenting with Unity and I have the COx-spheres back 😊

PR #828, opened 2016-12-03
comment out old RCS calculation

lamont-granquist:
MJ seems to be using the new stock RCS calculation, so there's no
reason to do the old calculation that I can see, and this fixes
the GUI debugging display to show the value MJ is really using.


lamont-granquist:
https://github.com/MuMech/MechJeb2/pull/828/files?w=0 to see the diff without my editors aggressive auto-fixing of whitespace.

lamont-granquist:
(note that i didn't confirm that torqueRcs is actually right, i just noticed that the code has apparently started trusting it, and I don't think this changes any behavior)

sarbian:
If you comment that you stop MJ from computing rcsThrustAvailable and docking won't work.
The old value will be gone in the next patch

lamont-granquist:
Ah, derp, I thought that was getting calculated in the prior loop, didn't see it hiding in the one I commented out...

PR #835, opened 2016-12-12
Plz add atmospheric pressure (atm) into stats

fr33soul:
Plz add atmospheric pressure (atm) into stats..
We would like to use this measurement to auto-arm/deploy parachutes!

source: [https://en.wikipedia.org/wiki/Atmosphere_(unit)](url)


sarbian:
A) It s already here in the Misc category
B) Don't press random button in github UI. a PR is not an issue

PR #836, opened 2016-12-12
improve transfer optimization

lamont-granquist:
updates the optimizer that picks transfer burns to be more
accurate.

the biggest change here is in not fixing the ending UT of
the transfer and looking for the closest match at that
UT.  i could have taken the two orbits and instead called
KSPs functions to search for the closest distance.  since
we're using an N-dimensional optimizer, though, i just added
the ending UT as another variable.

since constraining the UT was likely constraining the deltaV
of the solutions it seemed prudent to add the deltaV in as a
quantity to minimize.  the factor of 10,000 was chosen through
a bit of playtesting -- 100,000 was clearly too high.

with these tweaks accuracy of the transfer orbits have improved
greatly and the algorithm can generally nail the intercept, often
with 10 dV or so saved over the best value on the porkchop plot.

there's still issues with interactions.  kerbin-to-moho transfers
often get mun enounters and are somewhat poor.  kerbin-to-jool is
often perturbed but the solution generally seems to wind up in the
inner joolean moons.  kerbin-to-dres, kerbin-to-eeloo and
kerbin-to-moho are usually spot on as long as the mun doesn't get
in the way.

it seems like it should be possible to use the existing work around
that bumps the solution to the next orbit and restarts the solver in
order to avoid encounters, but i don't quite have a perfectly clear
picture in my head of how to do that in a generic way.

more useful would probably be to include a button on the porkchop
planner to include the destination capture DV in the plot and
selection.

### DIFF
--- a/MechJeb2/Maneuver/TransferCalculator.cs
+++ b/MechJeb2/Maneuver/TransferCalculator.cs
@@ -6,293 +6,293 @@ using UnityEngine;
 
 namespace MuMech
 {
-	public class TransferCalculator
-	{
-		public int bestDate = 0;
-		public int bestDuration = 0;
-		public bool stop = false;
+    public class TransferCalculator
+    {
+        public int bestDate = 0;
+        public int bestDuration = 0;
+        public bool stop = false;
 
-		int pendingJobs;
+        int pendingJobs;
 
-		// Original parameters, only used to check if parameters have changed
-		public readonly Orbit originOrbit;
-		public readonly Orbit destinationOrbit;
+        // Original parameters, only used to check if parameters have changed
+        public readonly Orbit originOrbit;
+        public readonly Orbit destinationOrbit;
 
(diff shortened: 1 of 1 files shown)


sarbian:
👍

lamont-granquist:
Sorry for the massive whitespace cleanup as well.  I can split that up if you like.

https://github.com/MuMech/MechJeb2/pull/836/files?w=0

lol, or not... =)

lamont-granquist:
damn you're fast

sarbian:
You got me as I was reading my mails :)

sarbian:
10 days later I noticed that it did not compile...
Forgot to save a file ;)

lamont-granquist:
hmmm, that's weird, what have i been playing with?

lamont-granquist:
mono complier on mac looks like it must be missing a few checks, your changes look good :+1:

PR #838, opened 2016-12-13
add OnFixedUpdate() hook to WarpToUT() API

lamont-granquist:
add OnFixedUpdate() hook to WarpToUT() API

previously the WarpToUT() API required the caller to constantly
beat on WarpToUT() every FixedUpdate().

this places a bit of an undue burden on the caller which can lead
to the existing bug in the node executor:

https://github.com/MuMech/MechJeb2/blob/02efd919a5c805ff7c01f92d34c81de7546c37bd/MechJeb2/MechJebModuleNodeExecutor.cs#L135-L146

when core.attitude.attitudeAngleFromTarget() >= 1 then the first block
is not executed.  but when timeToNode < 600 the second block is also
not executed.

this happens when MJ is at maxwarp allowing the navball to spin prior to
hitting the realign check.  throughout this period it is only
occasionally calling WarpToUT when thant AngleFromTarget() < 1 check
passes.

for relatively short/slow warps timeToNode gets called and drops it out
warp, but for long/fast warps you can skip so far that 10 minutes isn't
enough lead time to start slowing down.

so...

this patch doesn't fix the NodeExecutor but instead fixes the
WarpController so that callers don't have to beat on WarpToUT()
constantly (although they still can) and WarpToUT remembers its
last setting and manages warping to the UT without needing to
be called again.

### DIFF
--- a/MechJeb2/MechJebModuleWarpController.cs
+++ b/MechJeb2/MechJebModuleWarpController.cs
@@ -69,7 +69,7 @@ namespace MuMech
             SetTimeWarpRate(lastAskedIndex, false);
         }
 
-        // Turn SAS on during regular warp for compatibility with PersistentRotation 
+        // Turn SAS on during regular warp for compatibility with PersistentRotation
         private void SetTimeWarpRate(int rateIndex, bool instant)
         {
             if (rateIndex != TimeWarp.CurrentRateIndex)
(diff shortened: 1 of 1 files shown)


lamont-granquist:
(this may actually be bad enough of a bug you'd want to cut another point release with it)

PR #839, opened 2016-12-14
Revert errant merge commit

lamont-granquist:
don't think you really meant to do this.  looks kind of like you merged dev into master and then master into dev.

### DIFF
--- a/MechJeb2/GoodingSolver.cs
+++ b/MechJeb2/GoodingSolver.cs
@@ -1,0 +1,465 @@
+using System;
+using UnityEngine;
+
+namespace MuMech {
+    public class GoodingSolver {
+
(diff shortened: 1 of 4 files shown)


sarbian:
I had a problem with my local copy and I guess I did not resolve it properly :(

lamont-granquist:
yeah, looked like derp, but i couldn't quite figure out what you were trying to do...

i think there is part of the git flow workflow where the process of cutting a release is often to merge dev into master, then tag and release and then immediately merge master into dev, which i think helps with the merge commits that wind up in master's history that aren't in dev, but i think you have to do that before dev starts accepting more commits, and i was probably thinking about it way too hard...

PR #847, opened 2016-12-28
let attitude control settle before autowarp

OliverPA77:
This way MJ no longer relies on the 'stop rotation by warping' cheat and also improves Persistent Rotation compatibility.


sarbian:
I had to revert it since you used the wrong branch. Can you redo the PR on the dev branch ? Thanks. I can't seems to fix it now.

PR #880, opened 2017-04-23
Improved burn time countdown fixes

Starwaster:
* If manual throttle limit set to 0 then mark text red instead of green (feedback for player of negative consequences)
* If throttle limited due to unstable ignition then always set LimitMode.UnstableIgnition instead of only setting it when throttle up was attempted.
* Don't apply  throttleLimit to stageAvgAccel if LimitMode.UnstableIgnition. (instead, check manual throttle limits and apply those so we can still calculate burn time and burn countdown)
* Put extra check for countdown of timeToNode <= 0. Also check for Infinity in NextNodeCountdown()


Starwaster:
CRAP don't merge! (I stuck it in master)

PR #884, opened 2017-05-07
flameout or 0 restarts engines don't need ullage

lamont-granquist:
they aren't gonna start so ullage is useless.


lamont-granquist:
https://github.com/MuMech/MechJeb2/pull/884/files?w=1 for without the whitespace fixes that my editor autofixes

lamont-granquist:
hmmm, except i wonder if -1 ignitions left means infinity...

lamont-granquist:
also should probably make sure that the engine is subject to ullage before ullaging it

PR #886, opened 2017-05-12
fix ullage again

lamont-granquist:
#884 broke it pretty completely.

stops "BUG: getting ullage from RealFuels casted to null, ullage status likely broken" logspam


lamont-granquist:
d4fd98c291f22c1cb1fa411bb5ce5474a8e538d3 has the 1-line fix, rest is whitespace

PR #891, opened 2017-06-01
make stable ullage stable

lamont-granquist:
its probably obvious that this should have been fixed this way.

however, i was considering vessels which had means of ullage and when i
was considering how far MJ should ullage until it tried to light the
engine i was like "why not _very_ stable, that's better right?"

but for something like a sounding rocket where it is as ullaged as its
ever going to get, you want to allow users to throttle with 'stable' as
well.  right now if the prop status falls from 'very stable' to 'stable'
before you hit the throttle it breaks your launch.

really for sounding rockets users should turn this feature off.

should also consider autodetecting sounding rockets (if there's no
other engines or RCS on the craft, it'll never ullage, so this whole
feature is a bit breaking if its checked).

### DIFF
--- a/MechJeb2/VesselState.cs
+++ b/MechJeb2/VesselState.cs
@@ -1274,7 +1274,7 @@ namespace MuMech
                     return;
                 }
 
-                if ((propellantStatus != "Nominal") && (propellantStatus != "Very Stable"))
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@sarbian i accidentally sent this against master and not dev -- because they're in sync i didn't get the usual mess of extra commits...

sarbian:
Fixed :)

PR #893, opened 2017-06-02
rework of node executor ullage patch

lamont-granquist:
🚧 DO NOT MERGE YET 🚧 

Looking for @Starwaster 's feedback

Basically what this does is add another API that the ThrottleController exposes which is the "fixed" throttle limit.  Basically all the throttle limiting code other than ullaging sets this.  Then there is the throttleLimit which is still the actually applied throttle limit.  When ignitions are being prevented, typically the fixedThrottleLimit would be 1.0 while the throttleLimit would be 0.0.

The Node Executor now doesn't need to know as much about throttling and ullaging, and just trusts the fixedThrottleLimit from the throttle controller.

I also moved the infinity check around so that its a bit more aggressive about not presenting infinite burntime calculations through any API.  If there's a reason why some consumers need that patched and not others its definitely not obvious, and it reads more correct to be more aggressive about preventing infinite burntimes.

### DIFF
--- a/MechJeb2/MechJebModuleNodeExecutor.cs
+++ b/MechJeb2/MechJebModuleNodeExecutor.cs
@@ -36,10 +36,6 @@ namespace MuMech
             double dV = node.GetBurnVector(orbit).magnitude;
             double halfBurnTIme;
             double burnTIme = BurnTime(dV, out halfBurnTIme);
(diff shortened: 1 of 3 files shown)


lamont-granquist:
closing because the base was master

PR #894, opened 2017-06-02
NodeExecutor ullage refactoring

lamont-granquist:
Basically what this does is add another API that the ThrottleController exposes which is the "fixed" throttle limit. Basically all the throttle limiting code other than ullaging sets this. Then there is the throttleLimit which is still the actually applied throttle limit. When ignitions are being prevented, typically the fixedThrottleLimit would be 1.0 while the throttleLimit would be 0.0.

The Node Executor now doesn't need to know as much about throttling and ullaging, and just trusts the fixedThrottleLimit from the throttle controller.

I also moved the infinity check around so that its a bit more aggressive about not presenting infinite burntime calculations through any API. If there's a reason why some consumers need that patched and not others its definitely not obvious, and it reads more correct to be more aggressive about preventing infinite burntimes.

### DIFF
--- a/MechJeb2/MechJebModuleNodeExecutor.cs
+++ b/MechJeb2/MechJebModuleNodeExecutor.cs
@@ -36,10 +36,6 @@ namespace MuMech
             double dV = node.GetBurnVector(orbit).magnitude;
             double halfBurnTIme;
             double burnTIme = BurnTime(dV, out halfBurnTIme);
-            if (double.IsInfinity(halfBurnTIme))
-            {
-                halfBurnTIme = 0.0;
-            }
             return GuiUtils.TimeToDHMS(node.UT - halfBurnTIme - vesselState.time);
         }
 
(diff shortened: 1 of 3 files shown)


Starwaster:
Sorry but there's still something wrong, or something NEW that's wrong.
'Keep limited throttle over x%' has to have authority over the other limiters (except for ullaging because you don't want to ignite if unstable...)

So if 'Limit Q' or 'Limit throttle to %' want to lower thrust percent below 'keep limited throttle' then it has to succeed in keeping the minimum throttle to whatever it's set to. 

That doesn't matter so much in stock but it's lethal to Real Fuels because the engines can shut down and then you need another ignition (which you might not have) and plus you risk vapor in the fuel lines due to unstable conditions.

Right now, as I type this, in launch after launch I'm watching the other throttle limiters reduce throttle percent below the minimum value set by 'keep limited throttle over' (even set to 95%) and they just keep lowering it until the engines shut down and fail.

lamont-granquist:
That might be a totally separate bug?

I tried to replicate that just by setting the max throttle limit to 20% and I couldn't kill the engine (with no minimum throttle limit set), so couldn't figure out how to replicate it.  Which engine are you using that has this behavior?

Starwaster:
I was using the Titan V from KWR but I can also do it with a Mainsail and some orange tanks.

Set the Q limiter to 500 pa and the minimum throttle limiter to 95%.  It should literally be impossible for the throttle to drop under 95% but it does. That's what's at issue, not that the engine shut down. It's not an engine issue.

Starwaster:
Reproduced using only stock KSP 1.3.0 + MechJeb build 704 

It shouldn't be possible for any throttle limiter to lower throttle below the minimum designated percent, except the unstable thrust one but as the screenshot below demonstrates, it is doing exactly that.

screenshot86

PR #897, opened 2017-06-08
be more assertive about minThrottle

lamont-granquist:
Needed to handle fixed and temp throttle limits.

### DIFF
--- a/MechJeb2/MechJebModuleThrustController.cs
+++ b/MechJeb2/MechJebModuleThrustController.cs
@@ -311,8 +311,9 @@ namespace MuMech
            (i.e. there may be a more limiting temp limit) */
         private void setFixedLimit(float limit, LimitMode mode)
         {
-            if (throttleLimit > limit)
+            if (throttleLimit > limit) {
                 throttleLimit = limit;
+            }
             throttleFixedLimit = limit;
             limiter = mode;
         }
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@Starwaster can you check this out and see if this fixes (or breaks) anything?

lamont-granquist:
i think i've got a replication case now and this does not fix it.

Starwaster:
So hold off?

lamont-granquist:
Okay, so I'm playing around with this, and it looks to me like this is working as design.

So what I mean by that is:

- manual throttle has authority -- if you set a minThrottle limit you can go under that by throttling down -- also any MJ autopilot-ey modules that set a lower throttle will be able to.

- if I set a max throttle of say 50% and a min throttle of 95% i'll effectively get a "max throttle 95%" setting.

- i haven't tested maxQ specifically (i've been flailing around in orbit with an AJ-10) but it should work the same way -- if you set min throttle to 95% then the max Q limit will be able to drop from 100% down to 95% at which point it should stop.

So @Starwaster you should test this.

I still don't see where this patch fixes bugs.  I'm still a bit confused.

Its also very obvious to me that "min throttle" does not mean MJ won't ever kill your engine due to feathering throttle for some other reason, its very definitely "keep limited throttle over..."  not "keep throttle over..."

lamont-granquist:
validated that max Q limiting seems to be working for me

Starwaster:
I never said anything about manual throttling; look at that screenshot again. I clearly don't have manual control and that should not be happening and it wasn't happening before.

lamont-granquist:
Yeah, I see, I did basically the same test on top of this patch and it looks fixed.  Set Q limit to something low and and the min throttle to 90% and it caps limiting at 90%.  If I set the Q limit to something  just under max Q i can see the Q limit kick on a split second before it drops low enough to kick in the min throttle cap which holds it at 90% again.

Starwaster:
I'll test when I can, I can't get to it right now.

Starwaster:
Looks good so far. I've run it through a few flights and orbital tests and
it seems to be behaving itself

PR #898, opened 2017-06-10
Fix launch negative inclination circularization bug

### LINKED ISSUE #857: #685 "Launch into plane of Target" (wrong inclination)
With a "launch into plane of target" the inclinaton was calculated *-1 by reaching apoapsis.
On Launch the inclination is calculated to 52.1°, thats ok. If the rocket reach the apoasis the inclination swaps to -52.1°!

lamont-granquist:
negative inclinations are always the expensive way to change
inclination, positive inclinations are always the cheap way, and
that is baked into the function we are calling.

closes #857

### DIFF
--- a/MechJeb2/MechJebModuleAscentAutopilot.cs
+++ b/MechJeb2/MechJebModuleAscentAutopilot.cs
@@ -79,7 +79,7 @@ namespace MuMech
         public double currentMaxAoA = 0;
 
         public double launchLatitude = 0 ;
-   
+
 
         public double tMinus
         {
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@sarbian way trivial

sarbian:
Indeed :)

lamont-granquist:
@sarbian something to consider is that this circularize + change inclination maneuver is not necessarily going to happen anywhere near the AN/DN and could screw up target tracking and mess up the LAN.  it should probably be optional and likely disabled by default...   possibly clip it automatically if it adds more than 20-50% to the DV of the burn...

PR #899, opened 2017-06-10
fix low inclination ascent wiggles

lamont-granquist:
when launching from Kennedy (28.608) to the moon (28.363) the
ascent guidance would start (correctly) by burning due East (can't
launch to lower than your launch latitude).

as the craft followed a great circle track, the latitude of the craft
would fall allowing smaller latitudes.  however this resulted in the
craft attempting an inclination change at the end of the burn by
turning north (the opposite of a southwards jog at the start of the
burn) which was both expensive and totally screwed up plane targeting.

this fixes this by just not allowing inclinations which are below the
launch latitude (also allowing inclinations which are below the current
orbital inclination which would be fine if someone winds up in that
situation through some kind of button mashing).

this makes south-jogs very hard to do, but it results in very precise 0.25
degree errors when launching to the moon from KSC in RSS.  scanning
through NK's videos from 1.1.3-era he seems to have done no better than
this and often as bad as 0.40+ when launch to the moon with a south
jog.  i think mostly his south-jogs were fighting with the old MJ ascent
guidance which did no great-circle tracking.

adding south jogs would be hard since you need to 'know' the south jog
happened and allow the permissible launch inclination to lower.  it
might be necessary at that point to allow unclamped launch inclinations
and let people jog south and then correct, but i had a hard time getting
any better results than just with the clamped inclination.

### DIFF
--- a/MechJeb2/MechJebModuleAscentAutopilot.cs
+++ b/MechJeb2/MechJebModuleAscentAutopilot.cs
@@ -207,7 +207,7 @@ namespace MuMech
             //during the vertical ascent we just thrust straight up at max throttle
             if (forceRoll)
             { // pre-align roll unless correctiveSteering is active as it would just interfere with that
-                double desiredHeading = OrbitalManeuverCalculator.HeadingForLaunchInclination(vessel, vesselState, desiredInclination);
+                double desiredHeading = OrbitalManeuverCalculator.HeadingForLaunchInclination(vessel, vesselState, desiredInclination, launchLatitude);
                 core.attitude.attitudeTo(desiredHeading, 90, verticalRoll, this);
             }
             else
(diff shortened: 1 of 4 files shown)


lamont-granquist:
Correct solution is probably something like the Unified Powered Flight Guidance (https://ntrs.nasa.gov/search.jsp?R=19740004402) used on the space-shuttle, for now this'll have to do.

sarbian:
Nice link

abortz:
Isn't this more a bug with Ascent Guidance attempting to launch into a target plane using only a launch time and inclination? If I want to launch to an inclination slightly lower than my launch inclination, and not any particular plane, the old behavior was perfectly good.

I have no experience with the Mechjeb code so no idea how hard this would be, but alternatively, when attempting to launch into a target plane with too low an inclination, simply clamp the Ascent Guidance *target* inclination to the launch inclination but still compute the launch time the same as before. That should get you the same "very precise 0.25 degree errors" for Moon launches out-of-the-box, but still allow for doglegs.

Btw, I've managed around 0.15 relative inclination on a direct-to-Moon launch using Mechjeb ascent, but it wasn't user-friendly at all. ;-)

lamont-granquist:
> "clamp the Ascent Guidance target inclination to the launch inclination but still compute the launch time the same as before. " 

i think that's what we're doing now?

you'll get 28.3626 in the desired inclination field in the autopilot, which i'm fairly certain is what gets used for launch time.

then 28.608 - 28.3626 = 0.2454 so ~0.25 is the best you can do for a rocket that only flies 'straight' (downrange in the inertial system) without a yaw steering program.

abortz:
Right now 28.3626 ends up in the desired inclination field, which is what leads to the problems downrange on a moon launch. If instead that was just set to 28.608 (clamped to launch latitude when you select "Launch into target plane") but you still launch at the same time as otherwise, then as you said, you'll get the "best" straight launch for the Moon, but the autopilot can go back to not caring about launch latitude. This way you can still manually set/change the desired inclination to <28.608 when you don't have a target plane and you'll get best-effort results, and you can still manipulate the inclination field during ascent to get doglegs when desired, e.g. direct-to-Moon flights.

Btw, I apologize if I'm misunderstanding what the patch does, as I haven't played with it in-game.

PR #901, opened 2017-06-11
New script module: Smart A.S.S.

je4d:
KSP 1.3 doesn't work for me right now, so I couldn't build & use the latest dev. Hence this is against the last commit that works against KSP v1.2.


lamont-granquist:
if you revert b738cf7d30fe6f9b5f865b7e649972dcda48bcdf you can play with HEAD of dev on 1.2.2

sarbian:
I need to improve my merge-fu...

PR #902, opened 2017-06-12
fix 1kN thruster suicide burn calc

lamont-granquist:
i'm not sure if real fuels changed its API so that this no longer works
correctly or if it was not correct in the first place -- but this
definitely gets minthrust/maxthrust very wrong for the 1kN thruster
in realism overhaul / real fuels.

also applied the thrust limiter w/o the ullage clipping to zero to the
max thrust, which is what i thought this problem was going to be in the
first place -- this still looks like more correct code here, but wasn't
the root of the bug i was seeing in the suicide calcs.

### DIFF
--- a/MechJeb2/VesselState.cs
+++ b/MechJeb2/VesselState.cs
@@ -143,7 +143,7 @@ namespace MuMech
         public float throttleLimit = 1;
         /* the fixed throttle limit (i.e. user limited in the GUI), does not include transient conditions as limiting to zero due to unstable propellants in RF */
         public float throttleFixedLimit = 1;
-        public double limitedMaxThrustAccel { get { return maxThrustAccel * throttleLimit + minThrustAccel * (1 - throttleLimit); } }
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@sarbian do you recall the use case you were trying to fix with the code i backed out here?  it definitely causes issues now, but i can't guess at what you were trying to fix...

sarbian:
I think the comment is clear "RealFuels engines reports as operational even when they are shutdown"
The engine report(ed) isOperational true even when not ignited.

I guess this is not longer the case ?

lamont-granquist:
well, shutdown RF engines are not operational, so yes no longer the case?

but there's other conditions where an engine which is activated could be failed:  vapor in the fuel lines, out of ignitions, Agathorn strikes again, etc.

but those conditions should be addressed via baking in better integration with RF:  https://github.com/MuMech/MechJeb2/blob/master/MechJeb2/VesselState.cs#L1232-L1254

could probably put that into a PartExtensions for isEngineReallyOperational() that checked for vapor, ignitions, and Agathornage and get better code reuse out of that...

PR #904, opened 2017-06-25
Ascent Guidance overhaul and closed loop guidance

lamont-granquist:
- breaks up Ascent Guidance extracting out features like tracking heading and AoA from pitch control.
- implements a sub-menu for different ascent profile styles
- implements a "stock style Gravity Turn":  pitching over hard, burning with insane TWR up to 45k or so, raising the Apo there up to the target altitude, then coasting around half of Kerbin and circularizing)
- implements the pitch program from an Atlas-Centaur/Surveyor Powered Explicit Guidance for RSS/RO style ascents


lamont-granquist:
Got all kinds of buggy sharp edges right now. 

Somewhat crappy video of the PEG mode in RO/RSS: https://vimeo.com/223045581

pianojosh:
This is obviously an edge case for most uses, but for launching to rendezvous, it could become necessary to insert at non-zero vertical velocity. Would it be possible to add the ability to insert not at perigee?

lamont-granquist:
yeah there's slots for it in the algorithm all marked "FIXME: assumes periapsis insertion" so rdT needs to be the vertical velocity at cutoff and vT and hT need to be fixed for velocity and angular momentum at cutoff.  ideally it should be possible to just target the moon and press a button and get a direct ascent launch, which is the problem that the PEG algorithm was designed to solve (for surveyor missions) -- same goes for orbital rendezvous or hitting precise orbital contracts.  would probably need the yaw steering law for that, along with some upgrades to the launch timing algorithm that MJ uses.

pianojosh:
@lamont-granquist One nitpick I've got is that sometimes really long numbers will be displayed in the ascent path details window making it spread wide enough to fill the whole screen. It seems to happen very quickly, so I haven't been able to capture a screenshot when it happens. It's usually during the pitch program before guidance has converged.

Would it be possible to display numbers in scientific notation with rounding if they are outside the 10^-3...10^3 range?

lamont-granquist:
@sarbian i think its time to smash merge and move forwards.   this has the two RO fixes as well.

sarbian:
Is it ready to merge now ?

lamont-granquist:
yo dawg

lamont-granquist:
so this is just iteration #1, i'll open a new PR shortly with iteration #2 and try to expand it to 3-dimensional targeting and clean up the architectural issues this version has with staging.

PR #906, opened 2017-06-30
Further docking AP fixes

Starwaster:
* Better handling of lateral separation closure to address issue where MJ doesn't want to close the final half meter of distance between docking axes. This is done by slowing closure rate down the Z axis and increasing lateral axis closure rate.

### DIFF
--- a/MechJeb2/MechJebModuleDockingAutopilot.cs
+++ b/MechJeb2/MechJebModuleDockingAutopilot.cs
@@ -46,6 +46,8 @@ namespace MuMech
         Vector3d zAxis;
         public double zSep;
         public Vector3d lateralSep;
+        public double relativeZ;
+        public double relativeLateral;
 
         ITargetable lastTarget;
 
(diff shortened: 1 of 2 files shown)


Starwaster:
There further seems to be a discrepancy with MechJebModuleRCSController.SetTargetWorldVelocity() in that very low velocity changes are unable to keep up with the target. This is more readily apparent when very low velocity limits are manually set in the docking AP GUI, something like 0.01 m/s renders the AP incapable of keeping up with the target. 

This could be why the docking AP seems so reluctant to close the final lateral gap between the two docking port axes. The closer it gets the lower it wants to reduce closure rate. It wrongfully appears as though the docking AP has simply stopped trying to close, or as if it is targeting the wrong point in space. (a common yet incorrect assessment among MJ2 users on the forums)

sarbian:
It is most likely a problem with math done 1 frame in the past. It tries to dock with the last frame position or something like that.

Starwaster:
I thought about that but gave up on trying to trace the callbacks.

I guess that's why doubling the lateral closure rate works then. If the port is going to end up one frame past where it thinks it's going to be then effectively I'm moving it two frames worth

PR #910, opened 2017-07-14
Rover AP overhaul for the configurable stock wheels

BloodyRain2k:
I've noticed that the RoverAP became absolutely incapable of driving towards a target and actually stop in time, so it was time to figure out again how GitHub works, which I have to every single time I get back to this.

Anyways:
- WarpToDaylight is default off now, no idea why I made it default to on
- steering PID was adjusted to hopefully work better with the now different wheels
- averaging samples for the ETA was reduced a lot because it was just too large
- lots of changes to how it treats speed and turning at higher speeds, aka no longer jerking the steering to max while going 30m/s
- reminded it how2brake because it somehow forgot
- renamed the AP's "Follow" button to "Drive", though I should probably make it change depending on whenever or not you have selected a vessel, but it seems a little less confusing to me now
- changed the "Drive" button to no longer reset the waypoint index so it can be simply used as a Start/Stop button, instead alt+click now sets the waypoint index to the first waypoint
- alt+clicking the current waypoint in the waypoint list will now set the index to -1 meaning no active waypoint at all

While I've tested this build for a longer while now including having a rover go 120km on Minmus at 35m/s where the only issues were hill ramping with excessive landings would it probably not hurt if other people could also test vehicles to see if there might be other issues.

### DIFF
--- a/MechJeb2/MechJebModuleRoverController.cs
+++ b/MechJeb2/MechJebModuleRoverController.cs
@@ -63,7 +63,7 @@ namespace MuMech
 		public bool BrakeOnEnergyDepletion = false;
 		
         [ToggleInfoItem("Warp until Day if Depleted", InfoItem.Category.Rover), Persistent(pass = (int)Pass.Local)]
-		public bool WarpToDaylight = true;
+		public bool WarpToDaylight = false;
 		public bool waitingForDaylight = false;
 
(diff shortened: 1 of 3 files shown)


sarbian:
oops, forgot to merge. sorry for the delay.

PR #915, opened 2017-07-22
Add node merging to Maneuver Editor

abortz:
*Super* useful when doing standard burns that involve both in- and out-of-plane delta-V (e.g. changing inclination and periapsis/apoapsis for transfer to geostationary).

Example easy apogee kick into GSO:
1. Create maneuver to change inclination to 0 at highest AN/DN
2. Create maneuver to raise periapsis to 35786km at a fixed time of 0s after next node
3. Merge nodes. Done!

UI could be improved, and attempting to merge nodes that are far apart in time leads to totally useless results, as expected.


lamont-granquist:
yay!

PR #917, opened 2017-07-22
Also autostage a decoupler if it's a fairing

### LINKED ISSUE #916: Autostage won't stage past procedural fairings
This is because proc fairings have a decoupler module in them.

abortz:
To support procedural fairings. Fixes #916

### DIFF
--- a/MechJeb2/MechJebModuleStagingController.cs
+++ b/MechJeb2/MechJebModuleStagingController.cs
@@ -94,9 +94,9 @@ namespace MuMech
             if (HasStayingChutes(StageManager.CurrentStage - 1, vessel))
                 return;
 
-            //only fire decouplers to drop deactivated engines or tanks
+            //only fire decouplers to drop deactivated engines or tanks or fairings
             bool firesDecoupler = InverseStageFiresDecoupler(StageManager.CurrentStage - 1, vessel);
-            if (firesDecoupler && !InverseStageDecouplesDeactivatedEngineOrTank(StageManager.CurrentStage - 1, vessel))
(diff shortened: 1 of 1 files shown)


lamont-granquist:
So interstage fairings won't hit this logic because they'd decouple and drop active engines/tanks, so they're protected via the short-circuit here still?

abortz:
Right. Though you're right to be worried -- I find all of this autostage logic to be a little sketchy. It's trying to do an awful lot on very little explicit information.

lamont-granquist:
Yeah, I just had to validate it for myself.  An interstage payload fairing that separates was *not* staged with the proper upper stage payload fairing.  Although after staring at the code long enough I had convinced myself that this change was going to be buggy, but it worked perfectly...

Thanks, that has been a super annoying bug for ages...

I still have checking out your launch inclination PR on my TODO list i'm trying to burn down.

If you want another annoying RO/RSS issue to fix what do you think about releasing launch clamps automatically once thrust has hit 99% or so of available thrust and/or been basically flat for a second or something?  (i.e. whatever condition equals "engines are fully spooled up")

abortz:
Hmm, I hadn't thought much about interstage payload fairings, so kinda lucky it worked. Also, awesome idea re: launch clamps! Gonna clean up the autostage code and add that in a PR shortly.

lamont-granquist:
Actually I'm not sure I tried it where the fairing was staged ahead of the decoupler -- but that seems kind of a weird thing to do, so perhaps the decoupler dropping the engines is the thing that protects the interstage fairing from being dropped?  IDK, so it might still be buggy but it may not matter a lot?  but players do crazy things with their rockets in this game...

abortz:
Lol. If you want check out #920 I tried to clarify the logic a bit. Main thing is I decided/realized that dropping an inactive engine was a (relatively) high-priority goal, so fairings and autostage fairing settings aren't even a factor.

PR #918, opened 2017-07-23
When matching planes, clamp target inclination to launch latitude

abortz:
This is an alternate fix for PR #899. It is preserves the ability to do easy azimuth jogs and best-effort inclination launches, while defaulting to matching the best plane. cc @lamont-granquist

### DIFF
--- a/MechJeb2/MechJebModuleAscentAutopilot.cs
+++ b/MechJeb2/MechJebModuleAscentAutopilot.cs
@@ -350,7 +350,7 @@ namespace MuMech
         //
         protected void attitudeTo(double desiredPitch)
         {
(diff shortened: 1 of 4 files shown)


lamont-granquist:
I'm pretty certain you can't do this.  As the vessel follows a great circle track vesselState.latitude falls and then it becomes unclamped and the vessel will attempt a northwards jog in order to fix a too low inclination at the end of the burn, instead of a doing a southwards job at the start of the burn.  This just reverts the behavior to what #899 fixed.

abortz:
Just tested it again and it works great. The clamping of inclination here only takes place before launch, so nothing becomes "unclamped" as the vessel latitude falls. Jogs only occur if you manually type into the target inclination field in ascent guidance.

PR #920, opened 2017-08-02
Autostage launch clamps and cleanup/clarify autostage logic

abortz:
Also fixes unreported bug in autostaging where bouncing vehicle will trigger staging unexpectedly. Also fixes minor performance issues in a few misc. functions related to autostage checks.

### DIFF
--- a/MechJeb2/MechJebModuleStagingController.cs
+++ b/MechJeb2/MechJebModuleStagingController.cs
@@ -79,9 +79,9 @@ namespace MuMech
             if (!vessel.isActiveVessel)
                 return;
 
-            //if autostage enabled, and if we are not waiting on the pad, and if there are stages left,
+            //if autostage enabled, and if we've already staged at least once, and if there are stages left,
             //and if we are allowed to continue staging, and if we didn't just fire the previous stage
-            if (!vessel.LiftedOff() || StageManager.CurrentStage <= 0 || StageManager.CurrentStage <= autostageLimit
+            if (StageManager.CurrentStage == StageManager.StageCount || StageManager.CurrentStage <= 0 || StageManager.CurrentStage <= autostageLimit
                || vesselState.time - lastStageTime < autostagePostDelay)
                 return;
 
@@ -94,15 +94,19 @@ namespace MuMech
             if (HasStayingChutes(StageManager.CurrentStage - 1, vessel))
                 return;
 
-            //only fire decouplers to drop deactivated engines or tanks or fairings
-            bool firesDecoupler = InverseStageFiresDecoupler(StageManager.CurrentStage - 1, vessel);
-            if (firesDecoupler && !(InverseStageDecouplesDeactivatedEngineOrTank(StageManager.CurrentStage - 1, vessel) || HasFairing(StageManager.CurrentStage - 1, vessel)))
-                return;
+            //Always drop deactivated engines or tanks
+            if (!InverseStageDecouplesDeactivatedEngineOrTank(StageManager.CurrentStage - 1, vessel))
+            {
+                //only decouple fairings if the dynamic pressure and altitude conditions are respected
+                if ((core.vesselState.dynamicPressure > fairingMaxDynamicPressure || core.vesselState.altitudeASL < fairingMinAltitude) &&
+                    HasFairing(StageManager.CurrentStage - 1, vessel))
+                    return;
 
-            //only decouple fairings if the dynamic pressure and altitude conditions are respected
-            if ((core.vesselState.dynamicPressure > fairingMaxDynamicPressure || core.vesselState.altitudeASL < fairingMinAltitude) &&
-                HasFairing(StageManager.CurrentStage - 1, vessel))
-                return;
+                //only release launch clamps if we're at nearly full thrust
+                if (vesselState.thrustCurrent / vesselState.thrustAvailable < 1/1.2 &&
+                    InverseStageReleasesClamps(StageManager.CurrentStage - 1, vessel))
+                    return;
+            }
 
             //When we find that we're allowed to stage, start a countdown (with a
             //length given by autostagePreDelay) and only stage once that countdown finishes,
@@ -110,7 +114,7 @@ namespace MuMech
             {
                 if (vesselState.time - stageCountdownStart > autostagePreDelay)
                 {
-                    if (firesDecoupler)
+                    if (InverseStageFiresDecoupler(StageManager.CurrentStage - 1, vessel))
                     {
                         //if we decouple things, delay the next stage a bit to avoid exploding the debris
                         lastStageTime = vesselState.time;
@@ -200,6 +204,20 @@ namespace MuMech
             return false;
         }
 
+        //determine whether activating inverseStage will release launch clamps
+        public static bool InverseStageReleasesClamps(int inverseStage, Vessel v)
+        {
+            for (int i = 0; i < v.parts.Count; i++)
+            {
+                Part p = v.parts[i];
+                if (p.inverseStage == inverseStage && p.IsLaunchClamp())
+                {
+                    return true;
+                }
+            }
+            return false;
+        }
+
         //determine whether inverseStage sheds a dead engine
         public static bool InverseStageDecouplesDeactivatedEngineOrTank(int inverseStage, Vessel v)
         {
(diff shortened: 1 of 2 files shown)


MechJeb2/MechJebModuleStagingController.cs
```diff
@@ -79,9 +79,9 @@ public override void OnUpdate()
             if (!vessel.isActiveVessel)
                 return;
 
-            //if autostage enabled, and if we are not waiting on the pad, and if there are stages left,
+            //if autostage enabled, and if we've already staged at least once, and if there are stages left,
             //and if we are allowed to continue staging, and if we didn't just fire the previous stage
-            if (!vessel.LiftedOff() || StageManager.CurrentStage <= 0 || StageManager.CurrentStage <= autostageLimit
+            if (StageManager.CurrentStage == StageManager.StageCount || StageManager.CurrentStage <= 0 || StageManager.CurrentStage <= autostageLimit
```
lamont-granquist: so this relies on the fact that MJ has "KSP stages plus one" stages?  that'll probably work, and likely fixes "the MET bug" but seems a little side-effect-ey...

MechJeb2/MechJebModuleStagingController.cs
```diff
@@ -94,23 +94,27 @@ public override void OnUpdate()
...
+            //Always drop deactivated engines or tanks
+            if (!InverseStageDecouplesDeactivatedEngineOrTank(StageManager.CurrentStage - 1, vessel))
+            {
+                //only decouple fairings if the dynamic pressure and altitude conditions are respected
+                if ((core.vesselState.dynamicPressure > fairingMaxDynamicPressure || core.vesselState.altitudeASL < fairingMinAltitude) &&
+                    HasFairing(StageManager.CurrentStage - 1, vessel))
+                    return;
+
+                //only release launch clamps if we're at nearly full thrust
+                if (vesselState.thrustCurrent / vesselState.thrustAvailable < 1/1.2 &&
```
lamont-granquist: that likely isn't enough.  1/1.2 is 88% and 88% of a 1.05 SLT is 0.87, which will drop the rocket into the pad.  there should probably be a check here that SLT is > 1.00 at least, and likely what RO users want is to fully spool up while still attached to clamps to maximize thrust while still having the pumps active.  if you drop the clamps early then you waste dV on not-quite-fully-spooled-up engines, and some of the bigger/later engines spool up pretty slowly.   thats why i was thinking 98-99% -- although i'd also almost think it should be a tweakable in some menu someplace.

abortz: Yeah I was trying to avoid extra UI. There's still a pre-stage delay (defaults to 0.5s) which seems to be enough extra time for the remaining spooling, but I only tested on a few rockets.

I did initially use a test of SLT >1, maybe I should just go back to that? I switched because I worried about rockets with SRBs that don't have SLT >1 without them. Also, are people really launching rockets with 1.05 SLT??

lamont-granquist: One of my testing rockets for the PEG stuff has a single F-1 booster stage with about that much SLT and then 4-6 RL-10s that only produce about 0.25-0.35 TWR in the upper stage.  The actual Saturn-V i think was something along those lines.

The SLT clamp also has the added feature that if TestFlight hits you with a failure on the pad, then your clamps wouldn't release if you never made it past 1.00 -- although at some point you'll probably want a way to walk through the engines and determine that they're all green and not Agathorn'd before releasing the clamps.

Ah, and SRBs...   that seems like you'd want to stage the liquid engines, spool them up, then release the SRBs (and i think those still have /some/ spool up time?) and then release the clamps after SLT > 1.00?  because even with SRBs you never want to release clamps with SLT < 1.00

I'm wondering if the pre-stage delay is such a good idea with launch clamp staging?  Maybe you could short-circuit the setting in this case and set it to zero?  Otherwise it seems like we're gonna burn up people's SRBs on the pad or risk dropping them into the pad depending on how we push things around...

MechJeb2/MechJebModuleStagingController.cs
```diff
@@ -94,23 +94,27 @@ public override void OnUpdate()
...
+            }
 
             //When we find that we're allowed to stage, start a countdown (with a
             //length given by autostagePreDelay) and only stage once that countdown finishes,
             if (countingDown)
             {
                 if (vesselState.time - stageCountdownStart > autostagePreDelay)
                 {
-                    if (firesDecoupler)
+                    if (InverseStageFiresDecoupler(StageManager.CurrentStage - 1, vessel))
```
lamont-granquist: I'm just curious why the change here, because its not obvious without context (just don't have time right now to read the rest of the code)

abortz: It's just a result of the code reorg -- it's exactly the same logic.

MechJeb2/MechJebModuleStagingController.cs
abortz: Okay, I think I'm convinced SLT>1 is a fine check. But SLT of 1.05 at liftoff is still *really* low -- I think Saturn-V was always >1.15. I'm inclined to leave the delay because it still makes sense here -- gives a little extra spool up time which will reduce losses if pumps are still running -- and the default will probably work well for most rockets. For rockets where it doesn't, well, it's configurable. :-)

MechJeb2/PartExtensions.cs
```diff
@@ -102,31 +102,17 @@ public static bool IsSepratron(this Part p)
 
         public static bool IsEngine(this Part p)
         {
-            for (int i = 0; i < p.Modules.Count; i++)
-            {
-                PartModule m = p.Modules[i];
-                if (m is ModuleEngines) return true;
-            }
-            return false;
+            return p.FindModuleImplementing<ModuleEngines>() != null;
```
sarbian: Did you test the memory usage of the stock call ? Most of the time when I use loops kile those it is because the stock code allocate memory for each call.
Could you do the same PR without those changes and I ll check the memory uses of the stock call (again)

abortz: Ah, good point @sarbian -- will do. Btw, what are good ways to test memory allocations in KSP/Unity?

sarbian: Unity Profiler
+
KSP in dev mode ( http://forum.kerbalspaceprogram.com/index.php?/topic/102909-ksp-plugin-debugging-and-profiling-for-visual-studio-and-monodevelop-on-all-os/ ) 
+ 
pairs of Profiler.BeginSample / Profiler.EndSample
Like there 
https://github.com/MuMech/MechJeb2/blob/master/MechJeb2/DisplayModule.cs#L164-L169
+
MJ compiled with the ENABLE_PROFILER symbol (Profiler.BeginSample uses a Conditional attribute)

abortz: Awesome, thanks.

MechJeb2/MechJebModuleStagingController.cs
abortz: Having 2nd thoughts:
1. More real-world rockets than I thought have SLT<1 without SRBs (e.g. all max-payload Atlas V configurations with any SRBs)
2. There's no easy way to make the (fundamentally zero-lookahead, greedy) autostaging work with spooling up multiple sets of engines.
3. SRBs don't seem to have spool up time in RO anyway (?)
4. Checking % available thrust will actually do the right thing in every situation I can think of *without* an engine failure
5. Engine failures are probably too complicated to handle easily and correctly here (even checking SLT is no guarantee due to reduced performance/thrust failures)

So I think % thrust is the right check. I'll make it >1/1.1 (91%) to *guarantee* SLT>1 for simple 1st stages. That doesn't have a noticeable impact due to pre-stage delay but it's good for piece of mind I guess :-)

lamont-granquist: I wouldn't think too hard right this second about Engine failures, I was sort of throwing that out there as a thing to consider how it would get dealt with eventually.  Probably something needs to get added to VesselState to determine if all the active engines are healthy or not (the icon turns yellow/red, so presumably there's some state there MechJeb could peek at directly), then it would reduce to just checking a boolean condition here.

I think the road you're heading down is roughly what I thought about and found no clear obviously correct answers, which was leading me towards wanting it to be tweakable in the UI -- and I would agree UI sprawl should be avoided if possible, but I can't quite see the obviously correct answer.

abortz: Agreed re: no always correct answer. Just thought I'd document some of the issues/compromises for you and others. You should try it out with any rockets you think might be problematic -- I suspect you may be pleasantly surprised. I think this is good enough to ship myself...

abortz:
@sarbian i say shipit! :-) thoughts?

lamont-granquist:
i am certainly a fan of "release early, release often", YAGNI and minimum viable product.

PR #924, opened 2017-08-18
Add a "Quick Warp" mode into MJ WarpHelper & WarpController

zyayoung:
I added a **"Quick Warp" mode** into MJ WarpHelper.

Features:
- Speed up Auto-warp a lot!!!
- It will save you a lot of time
- A Toggle to turn On/Off the "Quick Warp" mode

I like Kerbal Space Program and MechJeb quite a lot. 
So I just want to make a little contribution. XD

I wish you could accept my request. Thanks


sarbian:
I have some doubts about the physic precision at such rates but why not.

PR #930, opened 2017-08-27
Add an AirplaneAutopilotModule

zyayoung:
It works really well with small aircrafts such as Aeris-3A (even at very high speed)
But not very well with large aircrafts at high speed.

**Features:**
- Altitude Hold
  - Vertival Speed limit
- Vertical Speed Hold
- Heading Hold
  - Roll Limit
- Roll Hold
- Speed Hold
- PID adjustable

Problems:
- Can not control large aircrafts(or those with bouncing wings) well
- Heading Hold needs to be optimized
- Can not use Speed Hold separately (cannot use SAS while using Speed Hold)(easy to fix)

Screenshots:

**Hold the aircraft close to the sea level**
ASLFast

**4x Physics warp capable**(not for big aircrafts)
4xFastAndHigh

**Accurate**
Accurate

[More screenshots](https://github.com/zyayoung/image-repository)

### DIFF
--- a/MechJeb2/MechJebCore.cs
+++ b/MechJeb2/MechJebCore.cs
@@ -41,6 +41,7 @@ namespace MuMech
         public MechJebModuleSolarPanelController solarpanel;
         public MechJebModuleLandingAutopilot landing;
         public MechJebModuleSettings settings;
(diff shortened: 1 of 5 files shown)


sarbian:
Amazing, thanks for the PR

PR #931, opened 2017-08-27
Format Code & Disable roll control on the ground

zyayoung:
- Format The Code
- disable roll control on the ground
  - may prevent aircraft from turning over
  - using autopilot on the ground is still not recommended
- Release thrust control when autopilot disengaged
- avoid #932
- optimize VerticalSpeedHold


zyayoung:
Using FlightCtrlState directly is much better.
I'm working on that now. -_-#
💩

zyayoung:
Now it use FlightCtrlState directly to control the aircraft.

sarbian:
Oops, I forgot to merge it. 
I have some doubts about 2 MJ modules using FlightCtrlState but this will do for now.

PR #935, opened 2017-09-14
Changed DriveCircularizationBurn to set launchLANDifference off of co…

DavidShade:
…re.target.orbit.LAN instead of recorder.markLAN.  This fixes launching into plane of target.


sarbian:
An other mystery solved ! 
Thanks :)

Issue #940, opened 2017-10-15
New Update auto land issue version 2.6.1


Ksp567:
![kerbal space program screenshot 2017 10 15 - 17 43 34 26](https://user-images.githubusercontent.com/32809619/31582544-31c5e6be-b1d2-11e7-87d4-6456670a07df.png)
Using the land at target feature puts you directly 180 degrees on the other side of the moon/planet. Doesn't matter what planet or moon it is. 
![kerbal space program screenshot 2017 10 15 - 17 43 28 12](https://user-images.githubusercontent.com/32809619/31582542-2cc7fa12-b1d2-11e7-9a18-7766012e12fc.png)

bcwinters:
I have begun noticing issues with MechJeb land at target since the latest KSP update (was working fine previously).

It may be related to this KSP code change:
"CelestialBody GetRelSurfacePosition now calling correct methods."

https://wiki.kerbalspaceprogram.com/wiki/1.3.1

I can see that this method is used to compute the adjustments needed to find the landing spot in at least one place within MechJeb code:
https://github.com/MuMech/MechJeb2/blob/9be3000e57a8140a40831ee3193b9939ab8d9afb/MechJeb2/LandingAutopilot/DeorbitBurn.cs#L46

flyright-c172:
Win10 KSP User. I am using MechJeb2-2.6.1.0 as my only mod. Since the upgrade to KSP V1.3.1.1891 I have had the issue described above. MechJeb still works fine except when trying to use landing guidance from an orbital spacecraft, MechJeb will try to land, unsuccessfully, on opposite side of Mun from the target. Same results at Minmus.
 Results are similar to the screenshots in the original post, though the spacecraft comes in at too high speed and crashes.

sarbian:
It should now be fixed in the dev build. I do not really have the tome to test so you ll have to see for yourself.
Thanks @bcwinters for the 1.3.1 change hint

flyright-c172:
sarbian, Thanks so much for looking at this.
I tested MechJeb Dev #745 landing guidance with a spacecraft orbiting Minmus. It behaves differently, but still not working. Target looks correct, but MechJeb remains in the "Moving to high orbit burn point" status for multiple orbits.
![screenshot370](https://user-images.githubusercontent.com/32875599/31684920-7bac0de0-b33e-11e7-8818-20240877a1da.png)

doktorjet:
[...]

PR #941, opened 2017-10-22
Dev

DavidShade:
Fixes for #940. Seems GetWorldSurfacePosition gives a "System" relative value instead of a body relative one.


sarbian:
Great, thank you for that !

duchenpaul:
@DavidShade @sarbian seems the Jenkins hasn't complied yet.

sarbian:
It seems the server lost its DNS and it can not contact github anymore. I will have a look tonight.

PR #943, opened 2017-11-01
Fix flight recorder gravity/steering losses calcs

aw1621107:
Patch from lamont on #RO:

    < lamont> steering and gravity losses don’t depend on the rotation
              of the planet below — think about an orbit over a spinning
              airless planet — so surfaceVelocity has to be incorrect

lamont also removed the gravityLosses -= <expr> because that was
compensating for surfaceVelocity being used in the initial gravityLosses
calculation instead of orbitalVelocity.


lamont-granquist:
👍

PR #946, opened 2017-11-05
Use FAR for VesselState aerodynamic updates if available

aw1621107:
Used delegates instead of MethodInfo.Invoke because the methods are called frequently and delegates supposedly perform better than MethodInfo.Invoke.

Don't know if the last two commits are strictly necessary, but figured might as well have MJ use whatever FAR is using for consistency.

### DIFF
--- a/MechJeb2/VesselState.cs
+++ b/MechJeb2/VesselState.cs
@@ -32,6 +32,8 @@ namespace MuMech
 
         public static bool isLoadedFAR = false;
         private delegate double FARVesselDelegate(Vessel v);
+        private static FARVesselDelegate FARVesselAoA;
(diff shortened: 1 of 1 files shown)


aw1621107:
I just realized, I'm not totally sure using just `FARVesselRefArea` for `areaDrag` was correct. Given it's described as `// Product of the drag surface area, drag coefficient and the physic multipliers`, and the original `areaDrag` calculation is 

```
areaDrag += p.DragCubes.AreaDrag * PhysicsGlobals.DragCubeMultiplier * PhysicsGlobals.DragMultiplier;
```

might a more correct value be

```
areaDrag = FARVesselRefArea(vessel) * dragCoef * PhysicsGobals.DragMultiplier;
```

?

PR #952, opened 2017-11-15
Add aerothermal heating constraint for fairing autostaging

aw1621107:
Allows fairings to be automatically jettisoned at a more realistic time.

Equation for aerothermal flux was taken from page 3-14 of the
Arianespace Vega User's Manual, April 2014 edition [0].

Default max flux was 1135 W/m^2 in both the Vega User's Manual and the
Falcon 9 User's Manual v2.0, so it seemed like a decent default.

    [0]: http://www.arianespace.com/wp-content/uploads/2015/09/Vega-Users-Manual_Issue-04_April-2014.pdf
    [1]: http://www.spacex.com/sites/spacex/files/falcon_9_users_guide_rev_2.0.pdf


aw1621107:
Probably should have asked this earlier, but should I add an aerothermal heating field to VesselState (or at least make it accessible to the user somewhere)? Think the user should be able to see what the current value is if they're also going to have MJ take some action based on it.

sarbian:
Yes, it might be a good idea

PR #958, opened 2017-11-26
some fixes to corrective steering

lamont-granquist:
it now correctly does some kind of corrective steering to alter the
flight path angle.  i do not know if its equivalent to the prior
algorithm.


Starwaster:
Sorry to say but no, this doesn't really do anything useful. Yes, I know that people complained in the past that the old corrective steering made their rockets wobble but it also performed a useful function of correcting deviations from the planned flight path angle. 

Now, the deviations just grow and grow and no attempt at correction ever happens.

PR #960, opened 2017-12-02
fix advanced transfer lowest dV calc

lamont-granquist:
there's probably better ways to approach the whole problem of
interplanetary transfers, but this tweak should un-break this
algorithm.

it tries to create a cost well with an equality constraint of
less than 100,000m for the periapsis and then minimizes the dV
of the transfer when it is within that cost well.

it seems to actually miss by 1,000,000s of meters, which is a bit
odd (maybe we're not integrating through the next planetocentric
orbit patch because if we miss the SOI entirely we may not have
one?), but it now seems to actually work rather than just be broken.


lamont-granquist:
still looks like its buggy, but this code seems to be operating correctly, the surrounding code simulating the burn seems to be buggy somehow...

PR #962, opened 2017-12-05
more transfer calculator fixes

### LINKED ISSUE #843: Advanced Transfert to Another Planet Issue
Hi, 
first I wish to thx for this great mod. Now the problem:
I try to achieve a Kerbin-Duna Transfert...
 - First I put the vessel in a CIRCULAR orbit around kerbin at 640km (for warping). 
 - Now, I use MechJeb to plan an Advanced Transfert to another planet:
    * Select Duna as a target
    * Open MechJeb and choose "advanced transfert...."
    * First the node planner issue a Warning "Initial Orbit can't be hyperbolic"......but initial orbit is not hyperbolic, it's circular.
    * Anyway, I ignore this warning, I click on a blue area of the porkchop windows. and create a node, then execute it.
    * Wait...for the transfert window....wait....wait....wait.....
    * When it reach the transfert window, at start the maneuver is executed as intended but when the planned orbit reach the point where it escape from Kerbin SOI, Duna is automatically deselected, and next calculation and execution of the transfert node are wrong and the vessel miss Duna. Even if I try to reselect Duna as fast as I can, it doesn't work.

KSP Version:  1.2.2.1622
Mods: 
- [x]Science! 5.4
- BetterBurn Time 1.5.3
- Gravity Turn Continued 1:1.7.6
- Kerbal Engineer Redux 1.1.2.8
- MechJeb (of course)  2.6.0.0
- Module Manager 2.7.5
- AutoAsparagus 2.2.5
- Smart Stage 2.9.5
(All by CKAN)

ps: hope it's clear, sorry for the poor quality of my english.

lamont-granquist:
the major point of this fix is actually the removal of the line

orbit.StartUT = t

which seems to cause issues.  that was being done so that
PatchedConics.CalculatePatch() could be called with orbit.StartUT
consistently, but i unrolled the loop so it gets called with 't'
first and then orbit.StartUT.  more code, but avoids the hack.

then there's some successive APIs that are called:

- getTruePositionAtUT (if we're not even in the correct solar transfer
orbit somehow)
- getRelativePositionAtUT (for the solar transfer orbit)
- getRelativePositionFromTrueAnomaly(0) (aka periapsis distance once we
  get into the target SOI)
- difference in dV once we're hitting the planet pretty much exactly

since the magnitude takes sucessive jumps up at every step the optimizer
should be able to handle is and continuously trickle down to the target.

i pinned the UT_arrival time again.  i don't think it makes a lot of
sense to try to optimize that since someone may have tried to pick a
target off of the porkchop plot and they should get more or less what
they pick.

did a bunch of whitespace cleanup as well.

closes #843

### DIFF
--- a/MechJeb2/Maneuver/OperationAdvancedTransfer.cs
+++ b/MechJeb2/Maneuver/OperationAdvancedTransfer.cs
@@ -4,139 +4,139 @@ using Object = UnityEngine.Object;
 
 namespace MuMech
 {
-	public class OperationAdvancedTransfer : Operation
-	{
-		enum Mode
-		{
-			LimitedTime,
-			Porkchop
-		}
-		static string[] modeNames = {"Limited time", "Porkchop selection"};
-		public override string getName() { return "advanced transfer to another planet";}
+    public class OperationAdvancedTransfer : Operation
+    {
+        enum Mode
+        {
+            LimitedTime,
+            Porkchop
+        }
+        static string[] modeNames = {"Limited time", "Porkchop selection"};
+        public override string getName() { return "advanced transfer to another planet";}
 
-		double minDepartureTime;
-		double minTransferTime;
-		double maxDepartureTime;
-		double maxTransferTime;
+        double minDepartureTime;
+        double minTransferTime;
+        double maxDepartureTime;
+        double maxTransferTime;
 
-		public EditableTime maxArrivalTime = new EditableTime();
+        public EditableTime maxArrivalTime = new EditableTime();
 
-		const double minSamplingStep = 12 * 3600;
+        const double minSamplingStep = 12 * 3600;
 
-		private Mode selectionMode = Mode.Porkchop;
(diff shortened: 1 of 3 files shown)


lamont-granquist:
Note that 843 isn't totally fixed, but the remaining bugginess is that the algorithm we're using for computing patched conics can miss interactions leaving the source system.  For example in stock a Mun interaction is entirely missed (I've Debug.Log'd it and it doesn't show up as a SOI change) and therefore the departure hyperbola is off, and ultimately there's no SOI encounter with the target.

...

I actually just thought of a solution to that problem as I was typing out that I had no idea what to do about that.  But we can detect that problem via not having an encounter with the destination SOI, and then we could apply the existing logic to move the maneuver forwards one parking orbit rotation and try again.

lamont-granquist:
^ that idea does not work, so IDK, this may be the best i can do for now.

PR #966, opened 2017-12-06
Fix target orbital inclination label color change again

aw1621107:
Looks like I screwed up #964 by overwriting the orbital inclination color change fix


sarbian:
Ah, sorry about that. I should not merge from work while doing 2 other things :)

PR #976, opened 2018-01-17
Docking AP Move To Start safeguard

Starwaster:
* Simplified and made safer the part of the docking AP where it is 'moving toward start' (if Z separation is inside of the safe zone then move BACK. Otherwise hold distance until we finish moving laterally)


Starwaster:
Meant for dev branch... one of these days I'll do a pull request where I haven't fcked something up....

PR #988, opened 2018-02-08
Use mcs over gmcs as default compiler

EoD:
mcs has been the recommended compiler since Mono 2.11:
http://www.mono-project.com/docs/about-mono/languages/csharp/

I assume that the usage of gmcs is only because the Makefile is rather old. This PR changes the default compiler to the combined mcs.


sarbian:
Now that mono ships a proper msbuild we could also just use that.

EoD:
The newer version of mono don't seem to ship gmcs anymore, hence this PR here. I have no opinion about msbuild vs mcs.

PR #991, opened 2018-02-12
fix some bugs and optimize FLARE in auto landing

zyayoung:
AutolandApproachState.FLARE <- This State did badly when landing at a high speed.
I don't think it is necessary to flare on propose. When the aircraft touches down at a low speed and low sink rate, it will automatically flare. Otherwise, my plane can't land if its speed is high.

My solution is to keep vertical speed control on until rollout, and adjust vertical speed target.

Now, it can land Aeris-3A smoothly ranging in speed from 50 to 200 m/s, and stop at the end of the runway.

other changes:
- increased a bit pid at low speed
- disable AltitudeHold when autoland enabled
- add a touch down speed TextBox.

Touch down speed textBox controls how fast the airplane touch the ground.
Approach speed controls the speed before final approach.
Maybe there are some mistakes in using the language, hope they can be fixed.

Sorry for my English -_-#

### DIFF
--- a/MechJeb2/MechJebModuleAirplaneAutopilot.cs
+++ b/MechJeb2/MechJebModuleAirplaneAutopilot.cs
@@ -147,7 +147,7 @@ namespace MuMech
             if (speed > 600)
                 return 0.5;
             else
-                return speed * -0.00125 + 1.25;
+                return speed * -0.0015 + 1.5;
         }
 
         void UpdatePID ()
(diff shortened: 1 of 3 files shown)


tekinozbek:
I'll this a try later tonight, but the change looks good to me.

By the way, I've been meaning to mention you that some heavier aircraft (there are a couple stock ones, like that airliner jet or whatever it is) have trouble banking left/right while maintaining the given target vertical speed. I didn't get a chance to debug this, but playing with the PID values don't seem to help all that much. Speeding up the aircraft helps, but then it tends to flare for way too long (which your changes would fix).

zyayoung:
Thanks. Maybe it is just too slow to control heavier aircrafts well, but the module itself does have some problems. The way Airplane Autopilot module work now is very simple, so i really hope it can be improved. I'll be very busy until June 2018. If the problem still exist, i'll try to fix it in July or August.

sarbian:
👍

PR #1014, opened 2018-04-09
Alglib update to 3.13.0

lamont-granquist:
bonus fix to add assemblies for KSP 1.4.x to the Makefile


lamont-granquist:
(i haven't actually tested this yet in the transfer planner, i can't imagine this patch doesn't work, but stranger shit has happened...)

sarbian:
Is it supposed to improve something or is it just to update?

lamont-granquist:
this is just to level it up.  i want to start playing around with the runge-kutta code and using levenberg-marquardt for root finding problems to solve trajectory optimization problems, and i don't want to be chasing down year old bugs or conflicts with the docs...

PR #1030, opened 2018-05-20
Landing tweaks

madman2003:
This is an attempt at tweaking the Landing Guidance to actually land first stage rockets without them falling over. In essence it tries not to turn the rocket above the ground to burn off last bit of horizontal velocity. Instead it turns on SmartASS to stabilize the last few seconds the landing, and SmartRcs after the landing is done.

This only fixes the unguided landing for me, the guided one still does silly things like 100% throttle just above ground with a Trust To Weight ratio of 30, which will throw the rocket into the sky and make it spin.

I'm looking for feedback/approval/etc before moving onto tweaking the guided landing.

### DIFF
--- a/MechJeb2/LandingAutopilot/FinalDescent.cs
+++ b/MechJeb2/LandingAutopilot/FinalDescent.cs
@@ -85,8 +85,13 @@ namespace MuMech
                 }
                 else
                 {
+                    float previous_trans_spd_act = core.thrust.trans_spd_act;
+
                     // last 200 meters:
                     core.thrust.trans_spd_act = -Mathf.Lerp(0, (float)Math.Sqrt((vesselState.limitedMaxThrustAccel - vesselState.localg) * 2 * 200) * 0.90F, (float)minalt / 200);
+
+                    // Sometimes during the descent we end up going up a bit due to overthrusting during breaking, avoid thrusting up even more and destabilizing the rocket
+                    core.thrust.trans_spd_act = (float)Math.Max(previous_trans_spd_act, core.thrust.trans_spd_act);
 
                     // take into account desired landing speed:
                     core.thrust.trans_spd_act = (float)Math.Min(-core.landing.touchdownSpeed, core.thrust.trans_spd_act);
--- a/MechJeb2/LandingAutopilot/FinalDescent.cs
+++ b/MechJeb2/LandingAutopilot/FinalDescent.cs
@@ -91,16 +91,12 @@ namespace MuMech
                     // take into account desired landing speed:
                     core.thrust.trans_spd_act = (float)Math.Min(-core.landing.touchdownSpeed, core.thrust.trans_spd_act);
 
-//                    core.thrust.tmode = MechJebModuleThrustController.TMode.KEEP_VERTICAL;
-//                    core.thrust.trans_kill_h = true;
-
-//                    if (Math.Abs(Vector3d.Angle(-vessel.surfaceVelocity, vesselState.up)) < 10)
                     if (vesselState.speedSurfaceHorizontal < 5)
                     {
                         // if we're falling more or less straight down, control vertical speed and 
                         // kill horizontal velocity
                         core.thrust.tmode = MechJebModuleThrustController.TMode.KEEP_VERTICAL;
-                        core.thrust.trans_kill_h = true;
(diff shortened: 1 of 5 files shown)


madman2003:
FYI, I will still need to tweak this a bit, because it still tries to push the rocket back in the air, but I'm obviously still looking for feedback

MechJeb2/LandingAutopilot/FinalDescent.cs
```diff
@@ -81,26 +83,42 @@ public override AutopilotStep Drive(FlightCtrlState s)
...
+                    // Sometimes during the descent we end up going up a bit due to overthrusting during breaking, avoid thrusting up even more and destabilizing the rocket
+                    core.thrust.trans_spd_act = (float)Math.Min(previous_trans_spd_act, core.thrust.trans_spd_act);
 
-//                    core.thrust.tmode = MechJebModuleThrustController.TMode.KEEP_VERTICAL;
-//                    core.thrust.trans_kill_h = true;
+                    // take into account desired landing speed:
+                    core.thrust.trans_spd_act = (float)Math.Max(core.landing.touchdownSpeed, core.thrust.trans_spd_act);
 
-//                    if (Math.Abs(Vector3d.Angle(-vessel.surfaceVelocity, vesselState.up)) < 10)
+                    // Prevent that we switch back from Vertical mode to KeepSurface mode
```
lamont-granquist: "Prevent switching back to KeepSurface mode from Vertical mode"

madman2003: I will update that once I get more feedback.

madman2003:
@sarbian Are you the main reviewer for this repo?

sarbian:
I do not currently have to time/will to look at anything. And I trust @lamont-granquist opinion.
So you are either confidant in your changes and I'll merge it or you are not.

lamont-granquist:
I don't quite understand the thrustAt200Meters bit, but a lot of this looks like it fixes edge conditions that I've definitely seen.  I'd lean towards merging.

lamont-granquist:
from the forums:

```
Hi, I recently had to revert back to dev 800 from dev 801 due to some severe issues with the new landing changes.

Scenario in 1.4.3 with dev 801:

Smallish Mun lander in circular 400k orbit around the Mun. No RCS on vessel
Changed Pe to 15k
at 15k, selected Land Anywhere
Just above the surface, the craft oriented itself vertically but still had a large horizontal speed (~10m/s). It slowly lost altitude while skimming over the surface for about a minute and crashed into the side of a hill, killing 7 tourists.
It seems the new code has some heavy reliance on RCS and does some auto enabling of attitude control. I'm not sure that's the best way to handle most cases. TBH, I always found the Land Anywhere code to be very good, with only occasionally minor hiccups. I would suggest:

Auto Enabling RCS is a mistake. In Career mode, it is common to make landers that to not need docking, so RCS is not added. Also, RCS should not play a role in Land Anywhere, unless the user enables the RCS button and RCS is present..
Main engines should always kill horizontal speed. The only thing that needed attention was the occasional flipping of the vessel excessively at the end to eliminate the last bit of horizontal velocity. Maybe raising the altitude a little and a gentler deceleration at the very end will fix that. If you detect that RCS button is pushed and RCS is actually present on the vessel, it can be used. Especially with Land at Target, where you are descending from a 500m height and you are a bit off, RCS can help on the way down
Thanks
```

MechJeb2/LandingAutopilot/FinalDescent.cs
```diff
@@ -81,26 +83,42 @@ public override AutopilotStep Drive(FlightCtrlState s)
...
+                    // When that happens the rocket will start tilting and end up falling over
                     if (vesselState.speedSurfaceHorizontal < 5)
+                    {
+                        forceVerticalMode = true;
+                    }
```
lamont-granquist: this is probably a bit flawed, the algorithm should really drive horizontal velocity to zero with main engines at 200m or so, with some clamps to make it not do silly spins hunting for a perfect zero speed -- then it can drop straight down once the horizontal speed is zero.

madman2003:
I see this got removed because it interfered with zero atmosphere landings. If and when I'm in the mood again to look into this, how to deal with this conflict of interest? I don't see myself finding a practical way of  something that works for both for tiny landers on the mun vs large rockets landing on kerbin. Should there be dedicated landing code?

PR #1056, opened 2018-10-14
Replace Hohmann Transfer with Bi-Impulsive Transfers

lamont-granquist:
The existing Hohmann Transfer planner only does circular-to-circular
coplanar transfer planning.

The new functionality does full Bi-Impulsive Transfer planning with
non-coplanar, non-elliptical orbits, including hyperbolic source
and target orbits.

It does this with a first pass through a simulated annealing algorithm
that does global optimization to try to find the lowest basin.  It then
hands that off to Levenburg-Marquardt to find the local minimum.

Since the simulated annealing algorithm uses a random search, it may
not find exactly the same answer every single time. The answers it finds
should be nearby and nearly the same dV but typically off by a few
orbital periods of either the source or destination or both (when the
search space has many possible orbits -- it should pick one of the
better ones, but not necessarily the absolute best).


lamont-granquist:
lol master != dev bites again

PR #1057, opened 2018-10-14
Replace Hohmann Transfer with Bi-Impulsive Transfers

lamont-granquist:
The existing Hohmann Transfer planner only does circular-to-circular
coplanar transfer planning.

The new functionality does full Bi-Impulsive Transfer planning with
non-coplanar, non-elliptical orbits, including hyperbolic source
and target orbits.

It does this with a first pass through a simulated annealing algorithm
that does global optimization to try to find the lowest basin.  It then
hands that off to Levenburg-Marquardt to find the local minimum.

Since the simulated annealing algorithm uses a random search, it may
not find exactly the same answer every single time. The answers it finds
should be nearby and nearly the same dV but typically off by a few
orbital periods of either the source or destination or both (when the
search space has many possible orbits -- it should pick one of the
better ones, but not necessarily the absolute best).

### DIFF
--- a/MechJeb2/GoodingSolver.cs
+++ b/MechJeb2/GoodingSolver.cs
@@ -18,6 +18,10 @@ namespace MuMech {
         public static void Solve(Vector3d R1, Vector3d V1, Vector3d R2, Vector3d V2, double tof, CelestialBody primary, int nrev, out Vector3d Vi, out Vector3d Vf) {
             /* most of this function lifted from https://www.mathworks.com/matlabcentral/fileexchange/39530-lambert-s-problem/content/glambert.m */
 
+            // if we don't catch this edge condition, the solver will spin forever (internal state will NaN and produce great sadness)
+            if ( tof == 0 )
+                throw new Exception( "MechJeb's Gooding Lambert Solver does not support zero time of flight (teleportation)" );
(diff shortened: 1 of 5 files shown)


lamont-granquist:
NOTE: Bi-Impusive transfers (two-burn) are not Bi-Elliptic transfers.  Bi-Elliptic transfers are actually Tri-Impulsive (three burn).

lamont-granquist:
screen shot 2018-10-14 at 8 38 09 pm

This is what the new "Hohmann" transfer planner window now looks like.

The drop down allows for scheduling the burn at a specific user-selected time using the familiar widget.  Otherwise the time of the burn is optimized.

The "intercept only" button is really the "mono-impulsive" option for doing flybys, impactors, launching nuclear missiles at things, where the second burn is of no consequence so is not included in the cost.

About all I can think of that is missing is a way to specify a transfer window period to search (minimum time in the future and/or maximum time in the future) and/or the maximum transfer time period to search -- plus a bit of code cleanup in #1058.  I think my UX brain threw an exception though and it is time to ship and playtest.

Gordon-Dry:
What does "fractional target period offset" mean?
(Using dev 808)

PR #1070, opened 2018-11-21
bring back simple coplanar transfer option

lamont-granquist:
The more cleverer transfer option does not work for everyone, so adding a checkbutton to get back to old style coplanar planning.


brianmcnett:
This is probably for the best. Co-planar transfers, while a specialized case, are easier for some folks to comprehend.

PR #1077, opened 2018-12-11
FuelFlowSimulation fixes

lamont-granquist:
* supports multiple moduleengines in a part (i.e. engines with integrated verniers, particularly SSTU russian engines)
* backs out the prop.ignoreForIsp fix.  the code as it stands now is actually correct (and was correct) and given the correct thrust and the rate of propellant consumption it figures out the correct ISP (the higher mass loss from HTP means that the effective ISP will be lower).   i suspect the ignoreForIsp fix was added due to the integrated verniers throwing off the thrust calculation or something like that.
* fix for RO leaving bigger bits of residual SRB propellant in the ullage rockets tank even though the motor has flamed out (mumble mumble real thrust curve something?).

also pushes a lot of the crazier bits of logic into PartExtensions that analyze the ModuleEngine.

backs out some of the dicts and use of linq to make the code a bit clearer.  this probably makes it perform marginally worse in the short term, but i'd prefer to try to fix the perf issues higher up in the algorithm if at all possible rather than having lots of complicated hard to understand dict caches lying around.

~i think there's still a bit of a known bug here where the thrust is always the combined thrust at the start of the stage, which is not what PEG/PVG need, but does give the correct starting TWR.  the burntime of the stage should still be correct, but the ullage engines will throw off the dV and the ISP.  i believe that is a very old bug though (see the comment at the top of ComputeTimeStepDeltaV).~  nope, testing shows that merging ullage solids with an upper stage works correctly, which means there's still some voodoo in here which i don't understand (makes sense or i think people would have reported that as a bug).

### DIFF
--- a/MechJeb2/FuelFlowSimulation.cs
+++ b/MechJeb2/FuelFlowSimulation.cs
@@ -147,7 +147,7 @@ namespace MuMech
             //print("Finished stage " + simStage + " after " + step + " steps");
             if (step == maxSteps) throw new Exception("FuelFlowSimulation.SimulateStage reached max step count of " + maxSteps);
 
-            //Debug.Log("thrust = " + stats.startThrust + " ISP = " + stats.isp + " FuelFlow = " + ( stats.startMass - stats.endMass ) / stats.deltaTime * 1000 );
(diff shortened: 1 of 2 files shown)


sarbian:
I ll will have a look at the perf once you your are done with the changes. The dict you removed are more a result of my laziness to do some of the math than actual perf optimisation.
And I will have to clean up some of the 'old' Tuple to replace them with something readable...

PR #1087, opened 2018-12-23
Incorporate "MechJebAndEngineerForAll" style functionality

lamont-granquist:
This adds MJ to all command pods by default.

It is opt-out, similar to the NoNonRP0 trick in RP-0 but requires
creating the empty folder GameData/MechJebUseCommandPod.

It also disables itself if RealismOverhaul or RP-0 are present since
those handle their own setup, and also override when techs are
unlocked.

I think this will address the incessant fly-bys in the forum thread
by people who install MJ and then see no display and bash out a one
sentence bug report stating that MJ is broken, because they've forgotten
that the install they've been playing for the past X months had
MechJebAndEngineerForAll in it.

I suspect 90% of players don't use the command pod, or wouldn't use
the command pods if they knew about MechJebAndEngineerForAll.

I think this may mean that MJ needs to pick up a dep on ModuleManager
in CKAN.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,7 @@ package: build ${MECHJEBFILES}
 	cp -r Icons package/MechJeb2/
 	cp build/MechJeb2.dll package/MechJeb2/Plugins/
 	cp LICENSE.md README.md package/MechJeb2/
+	cp MechJebNoCommandPod.cfg package/MechJeb2/
 
 %.tar.gz:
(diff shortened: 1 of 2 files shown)


sarbian:
I am away for xmass but it looks OK from my phone

lamont-granquist:
Well, code can always be changed/fixed so lets see how it goes...

Issue #1094, opened 2018-12-24
Overhaul scripting module for Ascent Guidance


lamont-granquist:
The way that the scripting module duplicates all of the state of the ascent autopilot and then has serialization / deserialization in readModuleConfiguration() / writeModuleConfiguration() is not very maintainable as the underlying autopilots are updated.

The scripting module probably needs to use the ConfigNode pulled directly off/on the modules in question.  Then the problem, though, is wrapping this ConfigNode for the state of the actual scripting module itself.  If we can just have a ConfigNode object (or 2 or 3) hanging off the MechJebModuleScriptActionAscent class then the problem is solved.

May want to have a rethink about how I've been semi-randomly hanging config off of different modules.

PR #1103, opened 2018-12-30
fix scripted ascents

lamont-granquist:
this likely breaks any existing programs, but extends the state
configuration across all the necessary sub-modules in a way that is
not brittle

launch to rendezvous/plane is disabled for now since it looks like that
state needs to get migrated properly to the AscentAutopilot class.


lamont-granquist:
mostly fixes #1094

PR #1108, opened 2018-12-31
Work In Progress:  onVesselStandardModification

lamont-granquist:
Going down the road of having this KSP event fire off static analysis to create sub-parts and sub-modules list which can then be walked in a more targeted manner.


lamont-granquist:
i'm just not gonna get to this clearly

PR #1146, opened 2019-03-09
Fix wobbly dV display in the last stage

lamont-granquist:
We need to synchronize the stage stats into the solution, but that
synchronization only happens when kicking off an optimization.  That
means that the state that was last synchronized may have no resemblance
to the current state of the stage stats, because the user may have
arbitrarily been rearranging staging in the meantime.  So we also
need to track the time that we synchronized the stage stats.

One open question I have here is why we don't simply update the stage
stats into the solution on every tick.  I don't recall why I didn't
do that.  Might have something to do with updating information
mid-optimization and creating a race condition with the next optimizer
cycle needing a consistent view over the whole time its running?  That
seems a bit weird, but the prior solution is consumed by the running
optimizer.

### DIFF
--- a/MechJeb2/Pontryagin/PontryaginBase.cs
+++ b/MechJeb2/Pontryagin/PontryaginBase.cs
@@ -33,6 +33,7 @@ namespace MuMech {
             public int ksp_stage { get { return _ksp_stage; } }
             int _rocket_stage;
             public int rocket_stage { get { return _rocket_stage; } }
+            public double _synch_time;
 
             public bool complete_burn = false;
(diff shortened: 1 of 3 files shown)


sarbian:
I think the main concern was slower PC that would take more than 1 frame and garbage generation.

lamont-granquist:
Oh no, this is different -- way in the guts of the Pontryagain/PontryaginBase optimizer madness.  That gets run off in a proper Thread, not just a Coroutine because it can certainly take a lot longer than a tick, particularly when bootstrapping.

PR #1147, opened 2019-03-10
Non active vessels enhancement

jrodrigv:
Hi @sarbian , @lamont-granquist 

This is a PR with two main enhancements for non-active vessels.
- When a non-active vessel is executing an ascent with auto-stage activated, the focus will automatically change to the vessel when it is required to stage and the it will switch back to the previous selected vessel after staging.
- Full review of mentions to FlightGlobals usage to remove the constraint of ActiveVessel.
- Enhancements to MechJebModuleTargetController.cs so it can set targets on non-active vessels.
- New Actions added to MechjebCore so users can execute Landing at KSC and Landing Somewhere operations.

* You can ignore the changes done to gitignore, sln  and csproj.
* I plan to continue adding more stuff to this PR as I go.


lamont-granquist:
acknowledged.  i probably need to adjust to DST first before looking at this.

sarbian:
We can't choose which changes to merge with the PR so you should remove the gitignore and project changes from your commit.

jrodrigv:
Sure I will amend the PR.

jrodrigv:
PR amended. I have also merged the latest changes from the Dev branch

lamont-granquist:
you can squash, rebase and force push your branch instead of merge committing if you want to...

MechJeb2/MechJebModuleStagingController.cs
```diff
@@ -206,7 +251,7 @@ public static bool InverseStageDecouplesActiveOrIdleEngineOrTank(int inverseStag
             {
                 Part p = v.parts[i];
                 Part decoupledPart;
-                if (p.inverseStage == inverseStage && p.IsUnfiredDecoupler(out decoupledPart) &&
+                if (p?.inverseStage == inverseStage && p.IsUnfiredDecoupler(out decoupledPart) &&
```
sarbian: I have some doubt about the use of this operator given how Unity override the null check but I guess that at worse it will behave as if it is not there.

sarbian:
Sorry for the delay, I kept forgetting it was there...

PR #1148, opened 2019-03-14
Fix FuelFlowSimulation for ignoreForIsp fuels (again)

lamont-granquist:
Screen Shot 2019-03-13 at 5 41 49 PM

Note the slider in the VAB all the way across to get the vacuum ISP numbers.

Note also here that this is *CORRECT* behavior.

The RD-108 upper stage has a 308.0s ISP on paper (literally in the design specs in the historical records).  Once the HTP is added to turn the turbopumps, the effective ISP falls to ~299.82s, which is what MJ calculates.

The RD-108 config in this case marks the HTP as "ignoreForIsp" which means that the engine config uses the "308s" value, KSP reports the "308s" value in the right click menu off the running engine, and RF reports the "308s" value.  The 299.82s value is, however, what the engine has always delivered (and how KSP has actually burned the fuel and really behaved).  This makes it easier for people to compare to historical documents, but the UX makes it more difficult to compare engine-to-engine, but the lower ISP MJ calculated value is actually correct.

Alternatively if the "299.82s" value were put in the engine configs and the HTP propellant were *not* marked as "ignoreForIsp" that result would be shown consistently in the UI and would make it easier to compare performance engine-to-engine for actual rocket design and the numbers would match between MJ and the KSP/RF UIs.

### DIFF
--- a/MechJeb2/FuelFlowSimulation.cs
+++ b/MechJeb2/FuelFlowSimulation.cs
@@ -828,13 +828,34 @@ namespace MuMech
 
                     if (e != null && e.isEnabled)
                     {
+                        // thrust is correct.
                         Vector3d thrust;
+                        // note that isp and massFlowRate do not include ignoreForIsp fuels like HTP and so need to be fixed for effective isp and the
+                        // actual mdot of the rocket needs to be fixed to include HTP.
+                        //
+                        // IMHO: using ignoreForIsp is just wrong.  full stop.  engines should never set this and should set the correct effective isp in
+                        // the config for the engine.  makes everything simpler and fixes the UI to show the correct ISP.  going this direction we are going
+                        // to get bug reports that e.g. mechjeb is reporting 299s for an RD-108 when RF and KSP are displaying 308s everywhere.  which is not
+                        // going to be a bug at all.  as long as the thrust is correct, the effective isp is correct, and the fuel fractions are correct then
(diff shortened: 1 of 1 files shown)


lamont-granquist:
Screen Shot 2019-03-13 at 6 13 01 PM

The other point to be made here (and what convinced me this was an issue) is to calculate the ISP from the fuel flows, which is reasonably easy since they're given in mass flows in the RF display.

ISP with HTP included:

ISP = 918300 / ( 89.6920 + 214.3357 + 8.2940 ) / 9.80665 = 299.82079

ISP without HTP included:

ISP = 918300 / ( 89.6920 + 214.3357 ) / 9.80665 = 308.00002

And one final note on a possible conceptual stumbling block.  The HTP is not part of the rocket exhaust so it "feels correct" to omit it, but that is not correct when looking at the mdot value in the rocket equation.  That value is how fast mass is disappearing from the rocket.

Where it would affect the problem is when using the real measured value of the rocket exhaust.  Working backwards from the 299s ISP value would produce a value of the rocket exhaust which was too low (although it is the "effective" exhaust velocity and can be used fine in equations that call for it).  Or using the real value of the rocket exhaust with the HTP would produce a value of thrust which was too high for the fuel flows.  Again though this is mostly useful when working with the historical documents of the rocket which may have the real value of the measured out-the-nozzle exhaust velocity.

PR #1169, opened 2019-06-20
Retain throttle setting when controls deactivate

RCrockford:
Retain throttle setting when controls deactivate by transferring the current throttle setting to the player controls.
This avoids crashing back into the pad if the player has not throttled up prior to starting autopilot.

### DIFF
--- a/MechJeb2/MechJebCore.cs
+++ b/MechJeb2/MechJebCore.cs
@@ -54,6 +54,7 @@ namespace MuMech
         private Vessel controlledVessel; //keep track of which vessel we've added our onFlyByWire callback to
         public string version = "";
         private bool deactivateControl = false;
+        private float currentThrottle = 0f;  // To maintain throttle when controls deactivated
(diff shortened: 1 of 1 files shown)


MechJeb2/MechJebCore.cs
```diff
@@ -71,7 +72,13 @@ public bool DeactivateControl
             {
                 MechJebCore mj = vessel.GetMasterMechJeb();
                 if (mj != null)
+                {
+                    if (value && !mj.deactivateControl)
+                    {
+                        controlledVessel.ctrlState.mainThrottle = currentThrottle;
+                    }
```
lamont-granquist: you probably want to move this block to after the next line, otherwise you're checking mj.deactivateControl before you set it.

lamont-granquist: Oh no I got it, its edge triggering, nevermind...

sarbian:
Thanks :)

PR #1170, opened 2019-06-21
Update MechJeb2.csproj

RCrockford:
Updated to current working state.

### DIFF
--- a/MechJeb2/MechJeb2.csproj
+++ b/MechJeb2/MechJeb2.csproj
@@ -38,18 +38,18 @@
     <WarningLevel>4</WarningLevel>
     <PlatformTarget>AnyCPU</PlatformTarget>
     <Prefer32Bit>false</Prefer32Bit>
-    <LangVersion>5</LangVersion>
+    <LangVersion>7.3</LangVersion>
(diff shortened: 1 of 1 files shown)


sarbian:
DLL hints should be in your .csproj.user and not in the main .csproj
Removing the Dll would break the jenkins build
I can accept the file list changes and the C version update.

RCrockford:
Hint paths should be back to normal, sorry about that.

MechJeb2/MechJeb2.csproj
```diff
@@ -264,6 +233,7 @@
     <Compile Include="ScriptsModule\MechJebModuleScriptActionWhile.cs" />
     <Compile Include="ScriptsModule\MechJebModuleScriptActionsList.cs" />
     <Compile Include="ScriptsModule\MechJebModuleScriptCondition.cs" />
+    <Compile Include="Shepperd.cs" />
```
lamont-granquist: 👍

sarbian:
Ok, let's merge is as is.

FTI the project has more dll reference than you have because on Linux and OSX Unity has split its engine dll in multiple one and for some reason they did not do it on WIndows until a release later...

PR #1190, opened 2019-10-27
Localization Things are done!

tinygrox:
I hope there won't be any mistakes


sarbian:
I cleaned up the PR with an other commit appart from that it looks OK.
Thanks for that work

PR #1207, opened 2019-12-15
Localization stuff

tinygrox:
Improved simplified Chinese Translation.
New localizable text content has been added

### DIFF
--- a/Localization/en-us.cfg
+++ b/Localization/en-us.cfg
@@ -83,9 +83,9 @@ Localization
 
 		#MechJeb_adv_Preconditions3 = must select a target for the interplanetary transfer.
 
-		#MechJeb_adv_Preconditions4 = doesn't make sense to plot an interplanetary transfer from an orbit around 
+		#MechJeb_adv_Preconditions4 = doesn't make sense to plot an interplanetary transfer from an orbit around <<1>> .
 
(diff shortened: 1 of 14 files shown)


tinygrox:
By the way, is there a solution to localize those text in the "MechJebModuleInfoItems.cs"

image
image

lamont-granquist:
Offhand I'd think you'd need to put in localization identifiers like `ValueInfoItem("#TimeToNode", ...)` and then the code that handled that would need to call `Localizer.Format` on that before using the string, but that's just a guess without actually looking at the code.

tinygrox:
> Offhand I'd think you'd need to put in localization identifiers like `ValueInfoItem("#TimeToNode", ...)` and then the code that handled that would need to call `Localizer.Format` on that before using the string, but that's just a guess without actually looking at the code.

Are you saying that... Like this?
image
I've tried it before and it doesn't work.

lamont-granquist:
MechJebModuleCustomInfoWindow.cs (i think) would need a corresponding patch to support that.  Those C# attributes are consumed there.

tinygrox:
> MechJebModuleCustomInfoWindow.cs (i think) would need a corresponding patch to support that. Those C# attributes are consumed there.

Wow, it worked.Thank you for taking the time to answer this.

PR #1208, opened 2019-12-18
More localized text

tinygrox:


lamont-granquist:
I think you have line editing issues with this PR?

MechJebCore.cs (for example) shows every line as updating but I don't see the diff, so I think you've converted the lined endings.  You shouldn't need to do that because the .gitattributes file should be treating all `*.cs` files as text and checking them out appropriate for the local system and doing line ending conversions on commit.

tinygrox:
Oh, I uploaded the wrong file. Fix it right away

tinygrox:
I uploaded the wrong files again so I had to spend some more time.And now it should all be done.

lamont-granquist:
Okay, gonna squash and merge this though to collapse the commits down

You might still want to look at the Localization/en-us.cfg file since every line updated there, but I don't think that warrants blocking merging.

PR #1211, opened 2019-12-26
A minor Localization correction

tinygrox:


tinygrox:
image
It is capitalized here in files _en-us.cfg_  and _zh-cn.cfg_

PR #1213, opened 2019-12-29
Allocation Free Dormand Prince Integrator

lamont-granquist:
Replaces the RKF45 integrator I wrote with the Dormand-Prince method.

Should correctly implement FSAL.

Uses stackalloc and Span so should be entirely garbage-free (but
temporarily requires compiling with -unsafe because mono or unity
or something doesn't fully support C# 7.2 initialization of Span
structures to stackalloc regions).

Over 4x faster than the RKF45 implementation (most likely due to code
cleanup that I did, I doubt my test code exercised the garbage
collector, so those benefits are on top of this speedup)

Not using Bogacki–Shampine's method because I ultimately want the free
4th order interpolant out of this method, which BS5(4) does not offer.

TODO:

- support for events (for the ReentrySim)
- 4th order interpolant (to replace the crappy alglib interpolation
  I'm doing in PVG)

This will get it to be broadly equivalent to Matlab's ODE45.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ build: build/MechJeb2.dll
 build/%.dll: ${MECHJEBFILES}
 	mkdir -p build
 	${RESGEN2} -usesourcepath MechJeb2/Properties/Resources.resx build/Resources.resources
-	${MCS} -t:library -lib:"${MANAGED}" \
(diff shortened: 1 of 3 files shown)


lamont-granquist:
@sarbian i think the csproj file will need an `-unsafe` flag shoved into it somewhere for VS users that aren't going in through the Makefile.

PR #1232, opened 2020-02-14
fix make file in case there are spaces in ksp path

voglster:
It wouldn't build on Linux with default path, traced it to missing quotes in the makefile.


lamont-granquist:
makes sense.

PR #1253, opened 2020-04-06
Add a simplified and commented `TimeToPlane` algorithm

dzhu:
This change replaces the previous implementation of `TimeToPlane` with a new one that is simpler and more commented.

It also fixes a bug in the handling of `LANDifference`: because `LANDifference` was subtracted as the very last step in the angle computations, a positive value could cause the output of `TimeToPlane` to be negative, leading to an immediate launch even when that was not correct.

(I'm not sure `LANDifference` is being set in the most useful way to begin with; it seems to always be better to leave it at 0 than use the value MJ fills in after a launch. But that's a separate discussion.)

Also, there is now a special case for very low inclinations (<.01 degrees); since there's such a small difference between different launch times, the function now just returns 0. The old implementation always returned a launch point 90 degrees away when the inclination was exactly 0.

### DIFF
--- a/MechJeb2/MechJebModuleAscentAutopilot.cs
+++ b/MechJeb2/MechJebModuleAscentAutopilot.cs
@@ -640,28 +640,82 @@ namespace MuMech
         //Computes the time required for the given launch location to rotate under the target orbital plane.
         //If the latitude is too high for the launch location to ever actually rotate under the target plane,
         //returns the time of closest approach to the target plane.
-        //I have a wonderful proof of this formula which this comment is too short to contain.
         public static double TimeToPlane(double LANDifference, CelestialBody launchBody, double launchLatitude, double launchLongitude, Orbit target)
         {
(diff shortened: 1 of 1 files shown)


lamont-granquist:
I actually cleaned that one up months ago on a branch I never pulled due to other issues with it.

I just rescued that in PR #1260, can you look at that and see if it works?  Its a little more extensive than what you had, and I just use one of napier's analogies from spherical trig to solve the problem (which is why I kind of prefer that solution since its just using a prebaked spherical trig algorithm and doesn't require any exposition or proof).

lamont-granquist:
replaced by #1260

dzhu:
Oh neat, that's some cool math.

lamont-granquist:
Yep, mathematicians already solved that one, I just needed to teach myself enough spherical trig to figure out which formula was simplest to apply.

PR #1259, opened 2020-04-18
Create es-es.cfg

Fitiales:
Add spanish localization


lamont-granquist:
That's pretty cool.  I can even read a few of those.

sarbian:
Thanks!
Now if there was a non lazy French around...

PR #1260, opened 2020-04-18
Math cleanup

lamont-granquist:
- Cleans up and extracts the TimeToPlane algorithm using clearer spherical trig computation
- Cleans up the angleToPrograde calculation
- The orbitLAN value is now always the orbitalLAN, even when landed
- The celestialLongitude value is the celestial longitude of the current vessel position (formerly what orbitLAN was when landed)
- The Vector3 ProjectIntoPlane and Vector3d ProjectOnPlane methods have been consolidated into one Vector3d ProjectOnPlane method


sarbian:
I am surprised by all the Vector3 vs Vector3d that are still lurking in MJ. I though I cleaned most of those a while ago.
This look good to me. 

And the "I have a wonderful proof of this formula which this comment is too short to contain." always annoyed me anyway.

lamont-granquist:
Heh, I didn't actually test that after extracting it from my other branch...

PR #1264, opened 2020-04-29
Add the ability to execute Principia nodes

raichuP:
This change adds the ability to execute Principia nodes. It adds a new "Execute next Principia node" button in the Maneuver Planner and Node Editor windows, and a new mode in the Node Executor module that executes the burn right on the burn time instead of start time minus half burn time. 

It also changes auto time warp behavior as Principia node is fixed relative to the selected coordinate system, caused it's always moving on the navball, made realign to the node 600 seconds prior to the node execution pretty unusable. So that the time of realigning to the node attitude is changed to half burn time plus 30 seconds prior to the burn time, this should work fine with both stock and Principia.

### DIFF
--- a/MechJeb2/MechJebModuleNodeEditor.cs
+++ b/MechJeb2/MechJebModuleNodeEditor.cs
@@ -230,7 +230,7 @@ namespace MuMech
                         core.node.ExecuteOneNode(this);
                     }
 					
-					if (GUILayout.Button(Localizer.Format("Execute next Principia node")))
+					if (MechJebModuleGuidanceController.isLoadedPrincipia && GUILayout.Button(Localizer.Format("#MechJeb_NodeEd_button7")))//Execute next Principia node
 					{
                         core.node.ExecuteOnePNode(this);
                     }
--- a/MechJeb2/MechJebModuleNodeEditor.cs
+++ b/MechJeb2/MechJebModuleNodeEditor.cs
@@ -230,8 +230,8 @@ namespace MuMech
                         core.node.ExecuteOneNode(this);
                     }
 					
-					if (MechJebModuleGuidanceController.isLoadedPrincipia && GUILayout.Button(Localizer.Format("#MechJeb_NodeEd_button7")))//Execute next Principia node
-					{
+		    if (MechJebModuleGuidanceController.isLoadedPrincipia && GUILayout.Button(Localizer.Format("#MechJeb_NodeEd_button7")))//Execute next Principia node
+		    {
                         core.node.ExecuteOnePNode(this);
                     }
 
--- a/MechJeb2/MechJebModuleNodeEditor.cs
+++ b/MechJeb2/MechJebModuleNodeEditor.cs
@@ -229,6 +229,11 @@ namespace MuMech
                     {
                         core.node.ExecuteOneNode(this);
                     }
+					
+					if (GUILayout.Button(Localizer.Format("Execute next Principia node")))
+					{
+                        core.node.ExecuteOnePNode(this);
+                    }
 
                     if (vessel.patchedConicSolver.maneuverNodes.Count > 1)
                     {
--- a/MechJeb2/MechJebModuleNodeExecutor.cs
+++ b/MechJeb2/MechJebModuleNodeExecutor.cs
@@ -124,7 +124,7 @@ namespace MuMech
 
             double timeToNode = node.UT - vesselState.time;
             //(!double.IsInfinity(num) && num > 0.0 && num2 < num) || num2 <= 0.0
-            if ((!double.IsInfinity(halfBurnTime) && halfBurnTime > 0 && timeToNode < halfBurnTime) || timeToNode < 0)
+            if ((!double.IsInfinity(halfBurnTime) && halfBurnTime > 0 && timeToNode <= 0) || timeToNode < 0)
             {
                 burnTriggered = true;
                 if (!MuUtils.PhysicsRunning()) core.warp.MinimumWarp();
@@ -137,7 +137,7 @@ namespace MuMech
                 {
                     core.warp.WarpToUT(node.UT - halfBurnTime - leadTime);
                 }
-                else if (!MuUtils.PhysicsRunning() && core.attitude.attitudeAngleFromTarget() > 10 && timeToNode < 600)
+                else if (!MuUtils.PhysicsRunning() && core.attitude.attitudeAngleFromTarget() > 10 && timeToNode < halfBurnTime + 30)
                 {
                     //realign
                     core.warp.MinimumWarp();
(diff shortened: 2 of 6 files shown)


sarbian:
Does Principia place its node in a way that the burn should be executed at their time instead of the half burn time ?

MechJeb2/MechJebModuleNodeExecutor.cs
```diff
@@ -137,7 +156,7 @@ public override void OnFixedUpdate()
                 {
                     core.warp.WarpToUT(node.UT - halfBurnTime - leadTime);
                 }
-                else if (!MuUtils.PhysicsRunning() && core.attitude.attitudeAngleFromTarget() > 10 && timeToNode < 600)
+                else if (!MuUtils.PhysicsRunning() && core.attitude.attitudeAngleFromTarget() > 10 && timeToNode < halfBurnTime + 30)
```
lamont-granquist, line 159: that needs to remain 600 seconds or something large like that.  if you don't you can wind up flipped around 180 degrees  30 seconds before the burn possibly with a huge vessel that cannot slew around that fast, and this will affect all users (stock, RSS/RO, principia, etc).

MechJeb2/MechJebModuleNodeEditor.cs
```diff
@@ -229,6 +229,11 @@ protected override void WindowGUI(int windowID)
                     {
                         core.node.ExecuteOneNode(this);
                     }
+					
+		    if (GUILayout.Button(Localizer.Format("Execute next Principia node")))
+		    {
+                        core.node.ExecuteOnePNode(this);
+                    }
```
lamont-granquist, line 236: I'm not wild about throwing this button at stock players, or even non-principia users.  This should probably be gated by a check that principia is loaded (there's some code in the guidance controller that starts to do that, it should probably get moved somewhere like VesselState (similar to the FAR checks there), or something like that, and made available so that every module can check if principia is loaded or not.

MechJeb2/MechJebModuleNodeExecutor.cs
raichuP: How about make a conditional statement: 
```suggestion
                else if (!MuUtils.PhysicsRunning() && core.attitude.attitudeAngleFromTarget() > 90 && timeToNode < halfBurnTime + 120 || !MuUtils.PhysicsRunning() && core.attitude.attitudeAngleFromTarget() > 10 && timeToNode < halfBurnTime + 30)
```

Or it could be limited to Principia only:
```suggestion
                else if (!MuUtils.PhysicsRunning() && core.attitude.attitudeAngleFromTarget() > 10 && timeToNode < 600 || MechJebModuleGuidanceController.isLoadedPrincipia && !MuUtils.PhysicsRunning() && core.attitude.attitudeAngleFromTarget() > 10 && timeToNode < halfBurnTime + 30)```

File: MechJeb2/MechJebModuleNodeEditor.cs (hunk 2 above)
Comment (raichuP) [reply MechJeb2/MechJebModuleNodeEditor.cs:236]: Done! However, the location of the code that check if Principia is loaded remain unchanged. .

Comment (lamont-granquist) [reply MechJeb2/MechJebModuleNodeEditor.cs:236]: heh okay bit wonky having it live there but that lets me clean that up later.

File: MechJeb2/MechJebModuleNodeExecutor.cs [hunk 3]
```diff
@@ -124,11 +132,22 @@ public override void OnFixedUpdate()
...
+				}
+			}
+			else if (mode == Mode.ONE_PNODE)
+			{
+				if ((!double.IsInfinity(halfBurnTime) && halfBurnTime > 0 && timeToNode <= 0.0225) || timeToNode < 0)
+				{
+					burnTriggered = true;
+					if (!MuUtils.PhysicsRunning()) core.warp.MinimumWarp();
+				}
+			}
```
lamont-granquist, line 150: can you reset your editor to keep tabs as spaces and not convert?  i wonder if visual studio respects some kind of magic comment policy...

MechJeb2/MechJebModuleNodeExecutor.cs
lamont-granquist: I'm a little confused about why you need to change this.  I thought this was the code where if you are very far from the time of the burn, MJ warps not caring at all about orientation (because stock KSP likes be wonky and cause some kind of precession even though its physics model shouldn't precess AFAIK -- but Principia probably precesses correctly) and then once you hit 10 minutes it starts to care about orientation and will drop you out of warp and fix the orientation if it gets too large.  That should AFAIK still apply to Principia?  Or maybe I've just forgotten what my own code patch does...

raichuP: The problem is that with Principia's node would have some kind of precession and causes node's direction move on the navball so that within 10 minutes from burn time it will drop out of warp like every 20 seconds because the node's direction has moved pretty far. It also happens when the node's reference frame is different than the navball's reference frame and that's even worse. It could be annoying if the ship's electric charge are low and in the shadow of the planet.

lamont-granquist: Uh, you need to check a box in Principia.  Inertially fixed or something like that?

raichuP: Yeah, I’m aware of that. However, inertially fixed uses slightly more deltaV and not works if the navball’s reference frame is different than the node’s reference frame, which is sometimes used when planning complex trajectories.

lamont-granquist: can you throw some kind of comment to that effect, specifically that it is to work around cases where the "inertially fixed" checkbox is insufficient?  since i wasn't aware those use cases existed.

lamont-granquist: (oh and go with the limit-to-principia only case probably, keep the use cases totally separate.

raichuP: Uh, you mean like open a new issue?

lamont-granquist:
mostly merged this in #1271

i did back out the change to the 10 minute terminal guidance period for the warp to the node.

i know from experience that had to be added due to large vessels that needed to essentially pre-aligned and then warped to the node because they align very slowly, if you take that out i know you break things for some users.

talking to egg, his opinion is that "inertially fixed is inertially fixed" so it doesn't make any sense this wouldn't work.

if there's a bug around that, it should probably get reported to one of us.

meanwhile i'll see if the principia players on the discord can use this or not and based on reports there will try to fine tune it more.

lamont-granquist:
(and it will take more delta-V, but i think i don't super care about that -- i'd rather waste people's delta V a little bit than have to deal with complaints of people who miss their node execution time with slowly slewing craft)

lamont-granquist:
egg is going to look at setting the maneuver node to be aligned with the inertial orientation with the start of the burn even if you don't have "inertially fixed" checked as well.

PR #1265, opened 2020-04-30
Added Russian Localization

craftersmine:
Added Russian locale with ru.cfg file.


lamont-granquist:
Well that looks like russian to me...

PR #1269, opened 2020-05-14
Avoid index error in ascent GUI

haltwise:
This fixes an index range error that occurs during certain scene loading situations. I particularly tended to encounter it if the ascent GUI was open when using the KCT SPH airlaunch feature, but also sometimes when quickloading a VAB flight.


sarbian:
Thanks !

Issue #1059, opened 2018-10-14
Docking Pilot EPIC issue rollup


lamont-granquist:
[ ] - tweakable for min distance for gross horizontal adjustments (so it stays 100m away from the docking port until it is aligned and then comes straight in -- #234

lamont-granquist:
Could the rendezvous autopilot benefit from a linearized gravity Hill-Clohessy-Wiltshire rendezvous/lambert solver like in Conway+Prussing, Orbital Mechanics, section 8.5 on Linear Impulsive Rendezvous?

PR #1285, opened 2020-06-08
#1059 - adds a field for max closing velocity in Rendezvous Autopilot

felixscheffer:


MechJeb2/MechJebModuleRendezvousAutopilot.cs
```diff
@@ -12,6 +12,8 @@ public class MechJebModuleRendezvousAutopilot : ComputerModule
         public EditableDouble desiredDistance = 100;
         [Persistent(pass = (int)Pass.Global)]
         public EditableDouble maxPhasingOrbits = 5;
+        [Persistent(pass = (int)Pass.Global)]
+        public EditableDouble maxClosingSpeed = 100;
```
lamont-granquist, line 16: should the default for this be something a bit more reasonable like 20?

lamont-granquist:
looks good like this though anyway.

Issue #1058, opened 2018-10-14
Refine Bi-Impulsive Transfers


lamont-granquist:
- [ ] the "maxUT" search space needs to be user tweakable
- [ ] the "maxUT" initial guess for tight circular parking orbit to hyperbolic orbit is likely off -- we're going to always launch nearly immediately, while parking orbits dozens of periods in the future could be better -- but Lunar transfers also have this feature where you're going to get better transfers once a month but in practice we don't care about that.
- [ ] the "maxTT" initial guess for the max transfer time needs to be user tweakable
- [ ] the other callers of DeltaVAndTimeForHohmannTransfer should be converted to DeltaVAndTimeForBiImpulsiveAnnealed
- [ ] similarly DeltaVAndTimeForHohmannLambertTransfer should be converted to DeltaVAndTimeForBiImpulsiveAnnealed?

PR #1287, opened 2020-06-10
1058 - allow an offset in front of / behind the target in km

felixscheffer:
#1058 

can you confirm that this is what you mean by "allow an offset in front of / behind the target in km"?

If not, feel free to modify.


lamont-granquist:
Nope, this is a hard one, you have to modify the true anomaly

lamont-granquist:
yeah probably going to close this, its just difficult to do right.

PR #1293, opened 2020-06-12
fixes label for glideslope in Landing Autopilot

felixscheffer:


Localization/en-us.cfg
```diff
@@ -256,6 +256,7 @@ Localization
         #MechJeb_ApproAndLand_label11 = Target altitude: <<1>> m
         #MechJeb_ApproAndLand_label12 = Target vertical speed: <<1>> m/s
         #MechJeb_ApproAndLand_label13 = Target heading: <<1>>º
+        #MechJeb_ApproAndLand_label14 = Glide slope:
```
lamont-granquist, line 259: I'm slowly going through and replacing the numerical designations with better descriptions so its more obvious what `#MechJeb_ApproAndLand_GlideSlope` is from `#MechJeb_ApproAndLand_label14`.

lamont-granquist: (that's just for future reference)

felixscheffer: Thanks. That makes way more sense.

PR #1297, opened 2020-06-14
improves Autoland

felixscheffer:
1. moves Final Approach Point further away from the runway. Realistically it should be something like 10000m but 5000m already improves things.
2. Climbing into the glideslope is really bad. We should maintain the current altitude until we intercept the glideslope. Unless we are very low of course, then we need to climb to a safe altitude (1500m should be fine for KSC).
3. 15m/s max rate of descent is kind of low. Probably okay for aircrafts but for spaceplanes it's way too low. So I turned that into an input field so the user can customize the value.

### DIFF
--- a/Localization/en-us.cfg
+++ b/Localization/en-us.cfg
@@ -257,6 +257,7 @@ Localization
         #MechJeb_ApproAndLand_label12 = Target vertical speed: <<1>> m/s
         #MechJeb_ApproAndLand_label13 = Target heading: <<1>>º
         #MechJeb_ApproAndLand_label14 = Glide slope:
+        #MechJeb_ApproAndLand_MaxRateOfDescent = Max. rate of descent:
 
         #MechJeb_ApproAndLand_approachState1 = Proceeding to the initial approach point
         #MechJeb_ApproAndLand_approachState2 = Proceeding to the final approach point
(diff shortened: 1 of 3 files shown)


felixscheffer:
I think we also need some sort of AoA limiter in the Autoland (or Aircraft Autopilot?)(like in the Ascent Autopilot to prevent erratic movements. The autopilot tends to pitch up to 30° to stop the aircraft from descending which causes the aircraft to slow down dramatically and eventually stall. 

The autopilot should keep the aircraft horizontally (maybe slightly pitched up) and increase the speed.

There is already maximumSafeBankAngle. Maybe we should add a maximumSafePitchAngle like in a flight envelope protection.

felixscheffer:
After looking at the code I have to say the Autopilot code is kind of awkward. Heading hold uses yaw to change headings instead of roll and maximumSafeBankAngle (aka RollMax) is only effective when both HeadingHoldEnabled AND RollHoldEnabled are enabled (which is not the case for Autoland).

sarbian:
Have a look at the file history. At some point we had a contribution that added a few things but made controls and a couple of other things worse. 
I kept thinking about reverting some of it but I never took the time ...

lamont-granquist:
This looks fine as an obvious enough tweak though.  About all I can think of is that someone might mash the autoland button when there's some kind of hill in the way...  Stalls are much worse though.

PR #1298, opened 2020-06-14
Fix Chinese Localization

WC12366:
fix a bunch of Chinese Localizations

(I don't think the translating machine works well)

### DIFF
--- a/Localization/zh-cn.cfg
+++ b/Localization/zh-cn.cfg
@@ -1076,7 +1076,7 @@ Localization
 
         #MechJeb_Actiongroup_label1 = 切换动作组
 
-        #MechJeb_Ascent_hotStaging = 允许过热分级
+        #MechJeb_Ascent_hotStaging = 允许热分级
         #MechJeb_Ascent_dropSolids = Drop solids early
         #MechJeb_Ascent_label38 = 分离整流罩当:
         #MechJeb_Ascent_label39 = 动压
--- a/Localization/zh-cn.cfg
+++ b/Localization/zh-cn.cfg
@@ -904,7 +904,7 @@ Localization
 
     //Translatron
 
-        #MechJeb_Translatron_title = 自动悬停
+        #MechJeb_Translatron_title = 平动控制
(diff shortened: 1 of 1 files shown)


tinygrox:
Just this one? Come on, you can do more! There are plenty of them!

Localization/zh-cn.cfg
```diff
@@ -887,26 +885,26 @@ Localization
...
-        #MechJeb_LandingGuidance_Status16 = Doing deorbit burn.
+        #MechJeb_LandingGuidance_Status13 = 离轨燃烧完成: 正在等待正确的时机减速
+        #MechJeb_LandingGuidance_Status14 = 正在改变轨道倾角，DV: <<1>> m/s
+        #MechJeb_LandingGuidance_Status15 = 正在移动至改变倾角点
+        #MechJeb_LandingGuidance_Status16 = 正在进行离轨燃烧.
 
     //Translatron
 
-        #MechJeb_Translatron_title = 平动控制
+        #MechJeb_Translatron_title = 自动悬停
```
tinygrox, line 907: Why did you change this one? It is not wrong.
Translatron is combine by "Translate" and "-tron"(I guess). And we all know there is a word in physics called `translational kinetic energy`--`平动动能`. The affix `-tron` means instrument or device. 
So the first word come to my mind is "平动控制器", then it became this.

Localization/zh-cn.cfg
```diff
@@ -924,7 +922,7 @@ Localization
 
         #MechJeb_Utilities_label1 = 自动分级 <<1>> 启动
         #MechJeb_Utilities_label1_1 = 一次
-        #MechJeb_Utilities_label2 = 飞机安全裕度
+        #MechJeb_Utilities_label2 = 飞机安全速度
```
tinygrox, line 925: `safety margin` is not the `safety speed`. Come on, you should see the `en-us.cfg` first.

WC12366:
thanks for your suggestions @tinygrox

PR #1305, opened 2020-06-23
Autopilot overhaul

felixscheffer:
So this PR turned out to be much bigger than expected.

- use roll for turns instead of yaw
- adds operational limits for pitch up/down, roll and yaw
- removes the "max descent rate" field from Autoland because it's not needed anymore.
- uses a pitch PID instead of a vertical speed PID.
- adds vertical speed field to altitude hold mode
- adds PID info text (current state, target, error and PID signal) to let the user know what's going on

tested with Aeris 3A, Gull, Mallard, and one of my own designs.

### DIFF
--- a/MechJeb2/VesselState.cs
+++ b/MechJeb2/VesselState.cs
@@ -151,6 +151,31 @@ namespace MuMech
 
         public MovingAverage3d angularVelocityAvg = new MovingAverage3d(5);
 
+        // instantaneous values
+        public double currentPitch
+        {
+            get
+            {
+                return (rotationVesselSurface.eulerAngles.x > 180) ? (360.0 - rotationVesselSurface.eulerAngles.x) : -rotationVesselSurface.eulerAngles.x;
+            }
+        }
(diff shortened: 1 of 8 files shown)


MechJeb2/VesselState.cs
```diff
@@ -151,6 +151,31 @@ public enum UllageState {
 
         public MovingAverage3d angularVelocityAvg = new MovingAverage3d(5);
 
+        // instantaneous values
+        public double currentPitch
+        {
+            get
+            {
+                return (rotationVesselSurface.eulerAngles.x > 180) ? (360.0 - rotationVesselSurface.eulerAngles.x) : -rotationVesselSurface.eulerAngles.x;
```
lamont-granquist, line 159: MuUtils.Clamp180 for these?

lamont-granquist: Ah I see where that already existed, but if you do more cleanup you might want to tidy that up.

PR #1339, opened 2020-07-29
French localization added

vinix38:
Added a single cfg file for the french translation.
Some words (only a few) are not localized and are therefore not translated.
I have favored coherence with the stock game over scientific terminology when in conflict.
I made sure that translated words worked ok with a 100%  UI, however I did not test with other UI settings.


sarbian:
Merci d'avoir eu plus de volonté que moi !

PR #1350, opened 2020-08-09
update unit label on vertical speed in aircraft autopilot

Rogach:
Degree sign on "vertical speed" field is confusing, looks like meters/second will convey the meaning better.


sarbian:
Yes, it looks like a copy paste error.

PR #1352, opened 2020-08-11
multiple Autopilot / Autoland fixes

### LINKED ISSUE #1163: Aircraft autoland unusably unstable
## Description

The Autoland feature in Aircraft Approach & Autoland seems to be unusably unstable, tending towards wild oscillations in attitude and speed, especially in pitch.  Changing PID values in Attitude Adjustment panel have no effect; indeed, it seems clear that that Autoland is not using the usual SmartASS PID controller because if I click "RESET" in Attitude Adjustment while autoland is running it makes no effort whatsoever to auto-tune itself until I abort autoland and renable SmartASS.

Autoland seems to be using its own separate attitude controller (reasonable enough, given that it should probably be doing banked turns), but its control parameters are unsuitable, especially for large (Mk.3) aircraft with large rotational inertia.

## Steps to Reproduce

1. Load this example craft: [Island Hopper.craft](https://github.com/MuMech/MechJeb2/files/3221627/Island.Hopper.craft.txt)
2. Take off and fly manually to somewhere plausible to start an approach from.
3. Open Aircraft Approach & Autoland.
4. Turn on Navball Guidance.
5. Activate Autoland.

### Expected behaviour

Aircraft aligns itself with target marker, then flies wings level at the configured speed.

### Actual behaviour

Aircraft bucks wildly in pitch, or repeatedly rolls left and right, never making an effort to roll out of the turn until past the desired heading.  (Ironically, the only time I see good pitch stability is during these repeated S-turns.)

With some craft I have, the pitch oscillations will get worse and worse until the craft stalls.

### Desirable additional behaviour

SmartASS allows manual keyboard override, but for some reason Autoland does not.  It would be very good it if did, because then there would be some way to damp down these autopilot-induced oscillations without havnig to disengage Autoland completely.

### LINKED ISSUE #1342: Autopilot wobbles up and down after reaching Altitude Hold
## MechJeb Version (DEV RELEASE)
2.10.0.0-1001

## KSP Version
Stock 1.10.1.2939 de-de
Making History 1.10.1
Breaking Ground 1.5.1

## Description
After engaging Aircraft Autopilot and reaching the set Altitude Hold of 7000 meters the airplanes starts flying a parable instead of keeping the set target altitiude. Like a bullwipe it swinged up and could even increase to an amplitude of 300 to 500 meters. I was only able to avoid it by decreasing Vertical Speed to 1 m/s and then to 0 m/s.
https://youtu.be/ghMSoK4600s

After a couple of minutes flying at a stable altitude of 7000 meters I set Vertical Speed back to 10 m/s and increased the time warp which caused again a parable flight.
https://youtu.be/riv0F3MhrPw

In some cases the airplane stops climbing at 6000 meters while vertical speed turns into negative values. After a drop of 500 meters it starts climbing to 7000 meters and above and remains in a parable flight. All parameters of the airplane, like speed or thurst are fine. Hence I can't explain.

This also causes a significant waste of fuel as the airplanes goes down and up and down and up.

## Replication Case
I took KSC runway 27 and engaged Aircraft Autopilot right after lift off heading 180° and then turning to 265°, altidude hold set to 7000 meters, V/S to 20 m/s, bank angle 30° and horizontal speed hold set to 120 m/s. The airplane should be capable of all of these settings. After reaching 6500 meters I set V/S back to 10 m/s. After reaching 7000 meters the above described behavior occured. I did not change any other settings. Just played a little bit with V/S in order to try stopping the annoying parable flight.

## KSP.log
[KSP.log](https://github.com/MuMech/MechJeb2/files/5010286/KSP.log)

### LINKED ISSUE #1343: Disengaging Aircraft Autopilot causes a nose dive as SAS will not be turned on
## MechJeb Version (DEV RELEASE)
2.10.0.0-1001

## KSP Version
Stock 1.10.1.2939 de-de
Making History 1.10.1
Breaking Ground 1.5.1

## Description
Disengaging aircraft autopilot causes a nose dive as SAS will not be turned on.

## Replication Case
After flying a couple of miles with engaged autopilot I disengaged it to perform some flight maneuvers. But with disengaging it SAS will not be automatically turned on. As I am not able to turn SAS on while autopilot is still engaged SAS remains turned off in the moment I disengage the autopilot. This causes an instable flight with a altitude drop of roughly 750 meters. I suggest that SAS shall be engaged in the same moment aircraft autopilot is disengaged in cases all prerequisites for SAS are fulfilled.

## KSP.log
[KSP.log](https://github.com/MuMech/MechJeb2/files/5010286/KSP.log)

### LINKED ISSUE #1349: Aircraft Autopilot allows to set negative Vertical Speed values
## MechJeb Version (DEV RELEASE)
2.10.0.0-1001

## KSP Version
Stock 1.10.1.2939 de-de
Making History 1.10.1
Breaking Ground 1.5.1

## Description
Aircraft autopilot allows to set Vertical Speed Hold to negative values. Same for VS +/- in case Altitude Hold is enabled. Setting a negative value causes the airplane to decrease altitude even if Altitude Hold is set to a much higher height. Regardless of that I always question the reason why it should be possible to set negative values. Doesn't the label already indicate that the entered values counts for ascent and descent? Shouldn't it be just allowed to enter positive values? At least I would expect that and that would be my suggestion.

## Replication Case
n/a

## KSP.log
[KSP.log](https://github.com/MuMech/MechJeb2/files/5010286/KSP.log)

felixscheffer:
fixes #1349 
fixes #1343 
fixes #1342
fixes #1163
fixes unit for V/S (should be "m/s"; was "°")

### DIFF
--- a/MechJeb2/MechJebModuleAirplaneAutopilot.cs
+++ b/MechJeb2/MechJebModuleAirplaneAutopilot.cs
@@ -20,7 +20,7 @@ namespace MuMech
         public EditableDouble AccKp = 0.5, AccKi = 0.5, AccKd = 0.005;
 
         [Persistent(pass = (int)Pass.Local)]
-        public EditableDouble PitKp = 2.0, PitKi = 0.6, PitKd = 0.65;
+        public EditableDouble PitKp = 2.0, PitKi = 1.0, PitKd = 0.005;
 
         [Persistent (pass = (int)Pass.Local)]
         public EditableDouble RolKp = 0.5, RolKi = 0.02, RolKd = 0.5;
@@ -92,6 +92,7 @@ namespace MuMech
         {
             if (!enabled)
                 return;
+
             _spd = vesselState.speedSurface;
             SpeedHoldEnabled = true;
             core.thrust.users.Add (this);
@@ -102,6 +103,7 @@ namespace MuMech
         {
             if (!enabled)
                 return;
+
             SpeedHoldEnabled = false;
             core.thrust.users.Remove (this);
         }
@@ -110,6 +112,7 @@ namespace MuMech
         {
             if (!enabled)
                 return;
+
             VertSpeedHoldEnabled = true;
             PitchPIDController.Reset();
             initPitchController = true;
@@ -119,6 +122,7 @@ namespace MuMech
         {
             if (!enabled)
                 return;
+
             VertSpeedHoldEnabled = false;
         }
 
@@ -126,6 +130,7 @@ namespace MuMech
         {
             if (!enabled)
                 return;
+
             AltitudeHoldEnabled = true;
             if (VertSpeedHoldEnabled) {
                 PitchPIDController.Reset();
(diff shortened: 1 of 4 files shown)


MechJeb2/MechJebModuleAirplaneAutopilot.cs
```diff
@@ -138,26 +143,27 @@ public void DisableAltitudeHold ()
...
             if (reference < 2)
                 reference = 2;
+
             if (deltaAltitude > reference || deltaAltitude < -reference)
                 return deltaAltitude / reference;
             else if (deltaAltitude > 0.1)
-                return deltaAltitude * deltaAltitude / (reference * reference);
-            else if (deltaAltitude > -0.1)
-                return 0;
+                return Math.Max(0.05, deltaAltitude * deltaAltitude / (reference * reference));
```
zerg960, line 159: could use Math.Sign() & Math.Abs() and merge both branches together, for clarity.

PR #1357, opened 2020-08-15
small bugfixes in waypoint window

lampeh:
- waypoint routes tried to load in loading screen
- range exception when copying speeds
- WaypointIndex out of range when deleting the last and active wp
- routes were not saved after "delete"
- "clear" would not clear the list unless a wp was selected

some additional whitespace cleanup is also included


sarbian:
Thanks!

PR #1361, opened 2020-08-28
Fix copy_build.bat to correctly support paths with spaces

Wetmelon:
There was some inconsistent quotes.  In VS, the batch file is called with all arguments passed with double quotes wrapping, but they were once again wrapped in "" in later checks, which would cause paths to look like

`""C:\My Path With Spaces""`, which fails.


lamont-granquist:
usually i see this bug the other direction

PR #1371, opened 2020-09-12
Add BetterController to VS csproj

Wetmelon:


lamont-granquist:
lol i always forget, i gotta start using rider on mechjeb

PR #1376, opened 2020-09-28
Run auto-ullage until engines start producing thrust

siimav:
Some engines in RealFuels can have rather long spoolup times during which the propellant can become unstable again. This PR fixes the issue by always running ullage for 0.15s longer after it becomes stable or until at least one of the engines produces 1% of it's rated thrust.

### DIFF
--- a/MechJeb2/MechJebModuleThrustController.cs
+++ b/MechJeb2/MechJebModuleThrustController.cs
@@ -232,6 +232,7 @@ namespace MuMech
 
         protected bool tmode_changed = false;
 
+        protected double ullageUntil;
 
         public PIDController pid;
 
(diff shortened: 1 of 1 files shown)


MechJeb2/MechJebModuleThrustController.cs
```diff
@@ -890,6 +871,52 @@ static void MaxThrust(double[] x, ref double func, double[] grad, object obj)
...
+
+            if (vesselState.time >= ullageUntil) 
+                return;
+
+            // Continue ullaging until at least one engine has spooled up enough to produce >1% of it's rated thrust
+            if (isStable && vesselState.enginesWrappers.Where(e => e.engine.requestedThrottle > 0.01)
+                                                       .All(e => e.engine.currentThrottle < 0.01))
+            {
+                ullageUntil = vesselState.time + 0.15;
+            }
```
lamont-granquist, line 897: How about ullaging up until the engines hit 0.95 or something like that?  Be simpler to write and understand.

siimav: The issue with that would be that nuclear engines take another 20 seconds until the throttle goes from 1% to 95+%. That 1% figure is what RF itself uses for determining whether the engine is in pre-ignition spoolup state. After 1% is reached then the thrust will start increasing very rapidly.
Relevant code for reference: https://github.com/NathanKell/ModularFuelSystem/blob/master/Source/Engines/ModuleEnginesRF.cs#L368-L374

PR #1377, opened 2020-09-28
Csproj file updates

siimav:
* .net 4.7.2 is a lot more likely to come preinstalled than 4.7
* update compilers package to latest version
* do not copy all the referenced assemblies to the output directory
* change debug symbol type to portable


sarbian:
Strange, I was sure I had the copy already disabled...

PR #1391, opened 2020-11-28
Use control point direction for RCS thrust

lampeh:
Calculation of thrustVectorMaxThrottle for RCS used the static control part transform, not the switchable control point and would not enable the RCS throttle when the control point was set to "reverse".


sarbian:
Somehow I missed that PR...
Thanks and sorry for the delay.

PR #1400, opened 2021-01-30
Update and fixes to French localization

vinix38:
Translation of the parts section
Fix of uppercase letters with accents
Fix some typos


sarbian:
Merci d'être moins paresseux que moi :)

PR #1405, opened 2021-02-22
add missing IconName override in CustomInfoWindow

### LINKED ISSUE #1403: Toolbar buttons missing
Having an issue where buttons are missing from toolbar. Linuxgurugamer said issues with missing buttons should be reported here. As you can see I’ve unlocked delta V, but it’s not an option in the toolbar available buttons. 
image

Machine-Maker:
fixes #1403 
Tested, and the icons now show up again with correct names, icons, etc.


lamont-granquist:
nice!

PR #1479, opened 2021-07-26
Chinese translation improvement

apollonius76:
"(to get rid of) ullage" should be "沉底"
"converge" (mathematics) should be "收敛"


sarbian:
Thanks for the contribution.

PR #1510, opened 2022-01-15
.NET 4.8 + ALGLIB 3.18 + refactoring

unquietwiki:
Hey there. The landing module was being wonky, so I started looking at the code. I got it to .NET 4.8 like some of the other plugins; and any Win10/11 box is going to have that; and the Mac/Lin systems should be able to support that via Mono/Core. [ALGLIB got updated to 3.18](https://www.alglib.net/arcnews.php#date_27_10_2021). [Roslynator](https://github.com/JosefPihrt/Roslynator) helped me refactor the code to a C# 8 standard. Hopefully it should be easier to make additional fixes; no behavioral changes are made in this pull. Thanks!


sarbian:
I am sorry, but I will not merge this for several reasons:
- I **HATE** _var_. My resharper config is set to use explicit type. I do not want to hoover a variable to know if it is a double or a float. And with KSP code this is not a trivial issue.

- The ?. Null Conditional Operator is dangerous with Unity. Unity runtime [overrides](https://blog.unity.com/technology/custom-operator-should-we-keep-it) the equality operator, and this does not work with ?.  Any use of ?. with something that inherits form UnityEngine.Object with not work as intended. I did not check the whole diff but I saw a couple and if you ran Roslynator over the whole project it most likely did it for some Unity objects null checks.

- Reformatting the full project will make life harder for other contributor that are working on patch. If Lamont is working on any patch for the ascent guidance then it creates more work for him. 

- [This](https://github.com/MuMech/MechJeb2/pull/1510/files#diff-6996df7217935673399251f2588a34321977639e14f7c115bbf9ec8c85a6a5f4) kind of formatting does not help IMHO. I get why adding brackets for single liner can be good engineering practice, but we also have tools that make sure the indenting is ok. Adding a bunch of nearly empty lines does not make the code more readable.

- 4.8 vs 4.7 does not change anything. In the end the code runs within KSP Unity Mono runtime and this runtime is neither NET 4.7 or 4.8. I do not mind changing it but there is no real impact.

I have no objection with a patch that fixes the formatting or minor issue in the file that they need to alter. But a right-click "fix all issue in project" is not something I want to merge.

unquietwiki:
@sarbian Thanks for the feedback. I can attempt to do this again with the guidelines you have suggested, and break it up some. The **.editorconfig** is our friend here; it's trivial to enforce the rules you've laid out, and I can even document this in a CODING_PRACTICES file or section. Will that be acceptable?

unquietwiki:
https://www.reddit.com/r/unity/comments/s4zt7t/question_about_a_unity_practice_from_2014/ I was also curious about the thing you brought up with the == testing. It looks like there's an object method you can use in its place.

unquietwiki:
@sarbian Regarding your wishes & concerns, I have two new commits that I think we both can agree somewhat on.

- [Commit 1](https://github.com/unquietwiki/MechJeb2/commit/629836d487ccc72e566b06fed589a92972b6eaad): this one is more in line with everything you laid out here. It establishes the coding rules, updates the libraries, and makes the process more clear for folks like me.
- [Commit 2](https://github.com/unquietwiki/MechJeb2/commit/a05edef38cd05e6384a69a516550f2eec9593974): when I did it the way you were not happy with, that was the result of 200K+ warnings & 100K+ suggestions (actually gave me the idea to upgrade my CPU, based on that chugging). With the guidelines you stated, this was reduced to just over 500; really just invoking **readonly** on runtime constants, and ordering the type declarations.

Please let me know what you think, and I'll tack accordingly. Thanks!

sarbian:
Both commits work for me. The second one is a bit more far-reaching, but is easier to deal with when merging. 
Feel free to add the CODING_PRACTICES  if you want to. MJ docs is lacking an many ways...

unquietwiki:
@sarbian Hey, thanks for getting back to me on this. For now, I put the coding standards in the README; organized that to be clearer. I appreciate your feedback, and the chance to help out!

lamont-granquist:
(FWIW, i'm not working on anything right now)

lamont-granquist:
(I'm also not a fan of var unless the type is redundant `var foo = new Something()`, even though resharper has those inline annotations to add the type back)

lamont-granquist:
oh one thing to keep in mind is that some of the code is synch'd from another project i was working on, which is why codestyle is a bit weird and there's a lot of excessive stuff that doesn't really match up with actual requirements MJ has (the files where you found the nullable annotations)

sarbian:
>doubles that can be floats (which some Unity calls apparently support)
There may be some useless cast but most of the time double are used because the precision is required. The KSP coordinate system does not play well at all in float.

I would be surprised that an Alglib update fixed a MJ bugs however.

unquietwiki:
@sarbian the ALGLIB release did mention some optimized SIMD DLLs that can be used for public & commercial versions, and might be compatible with the work here. I've been of the mind that trying to speed up calculations would help avoid possible misses on events & timing (KSP appears to parallelize decently, but MechJeb hangs can crash/stall it).

lamont-granquist:
@unquietwiki yeah i've been primary chef-client maintainer now for a long, long time.

the alglib updates might have actually helped with the transfer planner and the nonlinear optimization problem that it needs to solve.  but i suspect the problem there is more that the optimization problem needs tuning or else it needs improvement.  i keep thinking about what to do about that, but i'd really like to see RO/RSS get onto KSP 1.12 because many of the RO/RSS bugs may be related to the single precision Quaternion.LookRotation floats in 1.10

i'm not sure if the SIMD DLLs will help that much in terms of stability.  if those are precompiled dlls they might be problematic due to the target cpu architecture and possibly Mac/Linux/Windows support issues.

unquietwiki:
(nods) What's the RO thing? I'd like to take a look; this plugin has been a
timesaver for my gameplay, and I want to give back.

lamont-granquist:
That's just that KSP itself uploading bits of the maneuver node internals to get rid of some remaining 32-bit floating point operations in 1.12 and RO needs to upgrade off 1.10 -- the fix is already in KSP but RO can't use it until its built on top of that.

unquietwiki:
@lamont-granquist I found https://github.com/KSP-RO/RP-0/wiki/Introduction-and-Overview , which is what I think you were talking about for the RO/RSS thing. That looks like a complete overhaul... not sure if I can dip my toes into that, but I can give them some pointers on this effort.

sarbian:
They know what they are doing. They do not need pointers.
RO/RSS is huge and it takes them time to switch to a new KSP version and test it.

unquietwiki:
Acknowledged, just trying to help out. Been playing this on-off since Beta; finally got around to being more serious about it. I also do IT for a living, a lot of it unassisted at larger shops; so it's ingrained in me to solve problems, potential or immediate.

Did have an idea for something I'll submit a commit here on if I get it right: the translator works on surface movement, but I lose too much fuel with it, so I need to add a bursting mode. Also, might have an idea on how to make the landings more precise: I realized that coming down on Duna from 100km ended up 30km from the target; but 60-70km was only 3-4km off; so there might be some math calculation at work that could help offset that.

PR #1514, opened 2022-01-31
Optimize VesselState

DRVeyl:
Stock fixed their Moment of Inertia and Angular Momentum calcs from VesselPrecalculate, so use those instead of recomputing
Cache the results for calling FAR's CalculateVesselAeroForces.  The environment does not change much from frame to frame, but the call is expensive.  Set thresholds on the change in parameters (velocity, altitude) and use the previous result if within bounds
TODO: thresholds declared as constant are fine for demonstration, but they should probably be moved to a config area.

### DIFF
--- a/MechJeb2/VesselState.cs
+++ b/MechJeb2/VesselState.cs
@@ -913,6 +913,32 @@ namespace MuMech
             GUILayout.EndVertical();
         }
 
+        //FARCalculateVesselAeroForces(vessel,out farForce,out farTorque,surfaceVelocity,altitudeASL);
+        private Vector3 lastFarForce, lastFarTorque;
+        private Vector3d lastSurfaceVelocity;
+        private double lastAltitudeASL;
+        private const double _dVelocitySqrThreshold = 100;
+        private const double _dVelocitySqrMinThreshold = 0.0001;
+        private const double _dAltitudeThreshold = 300;
+        private const float _fOrientationThreshold = 5;
+        private void CalculateVesselAeroForcesWithCache(Vessel v, out Vector3 farForce, out Vector3 farTorque, Vector3d surfaceVelocity, double altitudeASL) 
+        {
+            if ((lastSurfaceVelocity - surfaceVelocity).sqrMagnitude > _dVelocitySqrThreshold
+                || Math.Abs(altitudeASL - lastAltitudeASL) > _dAltitudeThreshold
+                || (surfaceVelocity.sqrMagnitude > _dVelocitySqrMinThreshold && Vector3.Angle(lastSurfaceVelocity, surfaceVelocity) > _fOrientationThreshold))
+            {
+                FARCalculateVesselAeroForces(v, out farForce, out farTorque, surfaceVelocity, altitudeASL);
+                lastSurfaceVelocity = surfaceVelocity;
+                lastAltitudeASL = altitudeASL;
+                lastFarForce = farForce;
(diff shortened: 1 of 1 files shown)


siimav:
Has been working fine for me so far.

MechJeb2/VesselState.cs
```diff
@@ -913,6 +913,32 @@ public void RCSTorque()
...
+        private double lastAltitudeASL;
+        private const double _dVelocitySqrThreshold = 100;
+        private const double _dVelocitySqrMinThreshold = 0.0001;
+        private const double _dAltitudeThreshold = 300;
+        private const float _fOrientationThreshold = 5;
+        private void CalculateVesselAeroForcesWithCache(Vessel v, out Vector3 farForce, out Vector3 farTorque, Vector3d surfaceVelocity, double altitudeASL) 
+        {
+            if ((lastSurfaceVelocity - surfaceVelocity).sqrMagnitude > _dVelocitySqrThreshold
+                || Math.Abs(altitudeASL - lastAltitudeASL) > _dAltitudeThreshold
+                || (surfaceVelocity.sqrMagnitude > _dVelocitySqrMinThreshold && Vector3.Angle(lastSurfaceVelocity, surfaceVelocity) > _fOrientationThreshold))
```
lamont-granquist, line 928: what if your surface velocity is just small and bouncing around Vector3d.zero?  consider some sort of hovering vehicle  for example.

it sort of feels like it would be better to have a check for the angle that wasn't constrained, then a check for the change in the magnitude that was expressed more as a percentage (or a squared percentage maybe), and then the change in altitude check.

DRVeyl: I initially didn't constrain the angle check.  The realized low surface velocity hovering around 0 was going to trigger the angle check probably too often, so constrained it to a min surface velocity.  (_dVelocitySqrMinThreshold = 0.0001 is a 0.01m/s threshold.)

If you'd rather unconstrained and trigger the recalculate more often when truly not moving but only perfectly hovering, that's also fine.

Could also add a check and new parameter for the up direction on the root part, instead of assessing an orientation change from the change in velocity?  Would that be better?

DRVeyl: Thinking a bit more, these are only the aero forces, this isn't the entire computation.  Force and torque at near-zero velocity are zero, so you don't care.  You do, however, care about change of AoA at velocity great enough to have force/torque.

`(surfaceVelocity.sqrMagnitude > _dVelocitySqrMinThreshold && Vector3.Angle(vessel.rootPart.transform.TransformDirection(Vector3.up), surfaceVelocity) - lastAoA > _fAoAThreshold))`

That instead?  dVelocitySqrMinThreshold ~= 1.  fAoAThreshold (in degrees) ~= 2?

DRVeyl:
Would appreciate it if I can stay out of the UI piece.  You all have permissions via the PR to make edits.  Also a good opportunity to make those thresholds configurable.

Also made a change based on the conversation with lamont.

sarbian:
Sure. Tell me when you feel this is ready to merge and I will add whatever UI is needed later

DRVeyl:
Pending yours or lamont's concerns with https://github.com/MuMech/MechJeb2/pull/1514/files#diff-484374288e281725b5cace54bd2465245c31eaef4ddb123cee3d551b67a42ad6R925-R930  I believe this is ready to merge.

PR #1515, opened 2022-02-01
Optimize Stage stats simulation

siimav:
Optimize Stage stats simulation by copying second FuelFlowSimulation from the first instead of initilizing them twice.
Output from the profiler: 
image
So instead of 2*0.81ms it's now 0.81+0.23ms.


lamont-granquist:
This honestly looks like it is going to cause a pile of hard to track down bugs.

Consider the use case of adding a new one of those Dicts to this and how you need to make sure to add it in all the right places.

The code here is already very difficult to maintain.

siimav:
> The code here is already very difficult to maintain.

I heard you're planning to rewrite this part of the code from scratch anyway so as an interim solution I'm thinking this should be fine.

PR #1516, opened 2022-02-03
More linting rules & some landing tweaks

unquietwiki:
In my quest to get the landing work better, I consolidated the landing stuff & explored what other fixes could be done. I updated the landing site list with the stuff from the Wiki; there were some old & DLC sites missing. I kept running into an issue wherein, for some reason, the circularize function would end up broken (everything else would work). Furthermore, I also attempted to create a burst mode for the attitude mode; but while I could get the UI button added, I couldn't get the logic to work, so I abandoned that effort for the time being.

@siimav asked me about the .NET change. I did try to roll it to 4.7.2 at his request; I also had extended conversations with @Lisias regarding how .NET stuff interacts with the Mono runtime. AFAIK, anything in the 4.x range is compiled to the same baseline; the fun part is things like what you all were saying about Unity nulls & calling dependencies. Funny enough, KSP is compiled against the Windows/.NET Mono profile, and not the .NET Standard (2.0) profile; even though it's a cross-platform game. Trying to switch over to the newer csproj format might help with managing this, flipping to other versions, etc; otherwise, it's a PITA.
 
I did land on a good set of analyzers & rules to work from. It took a lot of trial and error to figure out what would work right; even then, [my upgraded CPU (working on this stuff was a trigger for going from 6/12 to 16/32 cores) chugs on analysis](https://imgur.com/a/PsnyEZA). I tried to get rid of any inane rules; ALGLIB has a lot of issues that I wish I could get the devs there to investigate.
 
So, TLDR: this isn't a massive change, but makes other changes possible.


unquietwiki:
https://www.nuget.org/packages/alglib.net/ ALGLIB is a Nupkg.... that helps a lot. Switch-over to that was successful!

lamont-granquist:
This probably needs to be broken up into multiple different PRs.

I'm :+1: on moving alglib to nuget to make it easier to manage, but that needs to be tested against @sarbian's jenkins build job.  Doing that is probably worth just a PR on its own.

The consolidation of the landing logic makes it difficult to check what you've done, and that was deliberately broken up into subclasses awhile back and that seems like going the wrong direction in terms of design to consolidate things.  The consolidation PR should probably be done on its own as well if that should get done, then the actual landing code changes should be a separate PR.

I'd also encourage you to back off on the enthusiasm for the linting stuff, it makes me very nervous based on how much code I have copypasta'd between my own code that I've been messing around with and the existing code in MJ and what I'd like to do in order to extract stuff into standalone mods in order to break up the MJ monolith.

(I'd love to understand the whole .NET Standard/Framework/Core mess, but I come from a Unix background and I totally can't wrap my head around it all)

lamont-granquist:
Also note that alglib is all written in some domain specific language which the author transpiles into C#, C++, etc.  So your chances of getting the author to address code style issues in the C# code is roughly nil.

unquietwiki:
@lamont-granquist (nods) I can try to reduce this PR to at least the linters & the rules; and then submit a new one for landing. The only issue I forsee with the ALGLIB Nuget, would be packaging whatever you guys put on CKAN; in my build artifacts, it comes up as a separate lib & there's a much lighter MechJeb file alongside it, and it's nice to not worry about a huge auto-generated library like you said.

I've been trying out additional linting & experimenting on a separate code branch altogether. Maybe I could do a PR for that for you to look at? I've found that whatever tends to mess with that, is a real sign of where actual bugs in the logic are; as you pointed out with your attempts to keep your code consistent.

unquietwiki:
@lamont-granquist I reverted the majority of what you were worried about, and started off with the final set of analyzers & rules I'd like to work with. If you approve that, I can submit a 2nd one for the alglib nuget, and you can see if that works.

Lisias:
Hi.

I did a peek on the commit https://github.com/MuMech/MechJeb2/pull/1516/commits/fd18b1f619f54be126b2b2a1ca7f85c0eb4b2f69 and I ask you to consider what follows:

It's usually a bad idea to switch from `for (int i; <end criteria; ++i)` loops to anything else due performance. Linq and functional features are somewhat less efficient at runtime, and make use of the heap for temp data - overloading an already overloaded garbage collector (due some Unity idiosyncrasies).

Outside Update, FixedUpdate and related callbacks I don't usually care too much about these optimisations, but on things that are called on every frame, the losses pile up - it's not unusual to have rigs doing 120 FPS or more, what implies on the Update and LateUpdate being called 120 times per second. Not a good idea to allocate heap on these callbacks.

Lisias:
> 



> @lamont-granquist I reverted the majority of what you were worried about, and started off with the final set of analyzers & rules I'd like to work with. If you approve that, I can submit a 2nd one for the alglib nuget, and you can see if that works.

Whoops… What renders my last comment irrelevant! On my defense, when I started to type the comment you didn't reverted it yet!

Cheers.

unquietwiki:
No prob, and I appreciate the feedback. I've done C# for stuff on/off since
2004, and there are still a bunch of features I haven't used yet, let alone
Unity. But I find this interesting work!

lamont-granquist:
I'm very interested in the alglib to nuget stuff, and have been thinking about that for awhile, but it is writing a check on sarbian's time probably to get it incorporated into the build system (it'd be nice to be able to pull stuff from nuget and build against it -- in an ideal world KSP mods could be published as nuget packages and brought in to link against them and then shipped separately as packages as well)

unquietwiki:
@lamont-granquist That reminds me of the warp mod I was investigating linuxgameguru maintains; that's tied to another library, that's cross-linked with another library. There's a Unity nuget out there too, but I saw that the author threw his hands up and cancelled further development.

lamont-granquist:
Yeah that might not be such a good idea for continuing maintenance, that might just be me in my head shouting at gamers to grow up and learn to be software developers...

But pulling alglib in from nuget is still a good idea.

unquietwiki:
If you're wondering why I'm here in the first place, it's because I am a
gamer & and IT pro like yourself. Also strongly believe in eliminating
technical debt; if not to make our own lives easier in terms of supporting
users & taking the stress off. It's also damn near impossible to stay
current on dev trends; so at the very least some automated fixes solve that
& educates the coder base. And in theory, good comments & code docs make up
for large changes. This codebase in particular is extremely complex, but
missing a lot of IDE-friendly documentation.

TLDR: I like getting rid of bugs & making it easier to maintain stuff to
avoid future ones, even if it's a PITA to make that switch.

lamont-granquist:
Oh sorry that wasn't directed at you, that was at my idea of having KSP modules all pulling each other from nuget and cross linking and then publishing up to github and CKAN.

That may be something that only a bare handful of people in the KSP space could really deal with.  That may be rubbing way too much software development on the problem and doing something that only that one other KSP mod does.

I was just crapping on my own idea out loud.

unquietwiki:
@lamont-granquist no worries. A GitHub org with various members could accomplish what you're looking to do. It'd also be useful to come up with an agreed standard for this stuff: I already ran into one add-on where the guy is compiling on Linux/Mono for a 4.8 target, and can even call features I can't with VS 2022 4.8 target. You'd think with KSP2 coming out, this would be less useful an effort; but the game will be more or less community-driven going forward, so maybe it does make sense to harmonize it all?

sarbian:
My main concern with an external DLL is how KSP loads them. As long as only MJ needs alglib we are fine, but if any other mods add an incompatible version we are screwed. It is not a likely scenario for alglib, but I had to deal with that last year and I do not want to do it ever again.

unquietwiki:
@sarbian Thanks for the merge. On the dependency issue, we only need Alglib; I haven't seen any others in the code, and it's not presently relying on features from .NET Std/Core 2.1+.

Also, for you & @lamont-granquist , I did put together https://unquietwiki.com/en/TechFindings/CodeQuality based on the stuff I've been doing lately; as well as in the past. I need to merge in / deprecate an older article from 2 years ago, else its good.

unquietwiki:
@lamont-granquist on our discussion of coding issues, I finally completed work on the fix for the FTL mod. That was fun in trying to sort out; had to short-circuit a NaN power draw.

https://github.com/linuxgurugamer/FTLDriveContinued/pull/27

PR #1517, opened 2022-02-05
Cached deployable handler

DRVeyl:
Don't iterate every frame over every Part&PartModule.  Instead, discover the list of valid ones and maintain that cache.  Invalidate the cache in response to OnVesselModified
Bugfix: Don't abort ExtendAll and RetractAll mid-way when *any* part is found to be shielded.
Rewrite ExtendAll/RetractAll/ExtendingOrRetracting/AllRetracted to use the cache

### DIFF
--- a/MechJeb2/MechJebModuleDeployableController.cs
+++ b/MechJeb2/MechJebModuleDeployableController.cs
@@ -111,11 +111,11 @@ namespace MuMech
                 prev_autoDeploy = false;
             }
 
-            bool allRetracted = AllRetracted();
-            if (extended != !allRetracted)  //State changed?
-                buttonText=getButtonText(allRetracted ? DeployablePartState.RETRACTED : DeployablePartState.EXTENDED);
+            bool extendedThisPass = !AllRetracted();
+            if (extended != extendedThisPass)
+                buttonText = getButtonText(extendedThisPass ? DeployablePartState.EXTENDED : DeployablePartState.RETRACTED);
 
-            extended = !allRetracted;
+            extended = extendedThisPass;
         }
 
         protected bool ExtendingOrRetracting()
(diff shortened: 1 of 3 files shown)


MechJeb2/MechJebModuleDeployableController.cs
```diff
@@ -177,5 +137,20 @@ protected enum DeployablePartState
...
+                DiscoverDeployablePartModules();
+                GameEvents.onVesselWasModified.Add(OnVesselModified);
+            }
+        }
+        public override void OnDestroy()
+        {
+            if (HighLogic.LoadedSceneIsFlight) GameEvents.onVesselWasModified.Remove(OnVesselModified);
+            base.OnDestroy();
+        }
+        public void OnVesselModified(Vessel v) => DiscoverDeployablePartModules();
```
lamont-granquist, line 154: This should probably get pushed down into MechJebCore and ComputerModule so that all modules get a vessel modification callback.  Also, what about `onVesselStandardModification`?  It seems to be the "superclass" of all vessel change events.

DRVeyl: This event handler is only relevant to the DeployableController.  I don't think there's a reason for every PM to get a callback just in case they could want to do something.  Especially if thus far none of them do. 
 Pushing it into MJCore or ComputerModule will mean that MJ registers on behalf of many handlers that will just NOP, and the pattern for registering and de-registering is straightforward enough.

You could swap to onVesselStandardModification instead, looks like.  I haven't used it, but that doesn't mean a lot.

Note there's nothing stopping ComputerModule from also wanting to subscribe to that Event, even if this module does.  The same PartModule could even conceivably subscribe to the same Event multiple times.  (This happens often, but as a coding error rather than an intentional thing.)

DRVeyl: Implemented in 24bc1c6

PR #1523, opened 2022-02-19
Use pre-packaged ALGLIB via NuGet

unquietwiki:
@lamont-granquist this is the separate ALGLIB via NuGet commit we discussed previously. I've been using this build for a few days now; think the circularization may be wonky, but it's been wonky for me regardless. (shrugs)


lamont-granquist:
This is sort of up to sarbian I think.

I in general philosophically like the idea of using package managers.

There may be downsides though in that if the alglib code is packaged in MechJeb.dll there's no possible issues around diamond dependency conflicts, whereas if its packaged separately then it becomes an issue of which dll wins, and ABI breaks may make it not worth it.

unquietwiki:
@lamont-granquist Based on the other add-ons I've been investigating; it's highly unlikely I see an issue with taking this approach. All the other conflicts I've seen have been either with add-ons against themselves, depending on each other, or trying to Nuget a recently-discontinued Unity abstraction library. The only add-on I've seen that could even touch what this one does, is the semi-maintained slingshot calculator; and that had no apparent dependencies.

I started creating a [separate branch](https://github.com/unquietwiki/MechJeb2/tree/optimize) to try out fixes & updates that could be rolled-in via cherry-picks. I successfully put a time library to use for resolving some linter complaints & offering millisecond-timed resolution (vs approximate). Furthermore, I also identified three other math libraries; that can replace some math code here & maybe deprecate ALGLIB. If we wanted to maintain the "no Nuget" approach; I would presume either of these changes would "weigh less" in added code than ALGLIB does now.

sarbian:
I am not in favor of using a separate DLL pour alglib. As I explained in the other PR DLL loading in KSP is a mess I had to deal with earlier and will not deal with again.
If you do not want to deal with the warning generated by the Alglib code then add a CAExcludePath to the project config.

Also, create separate branch for each things you do. I will not merge a PR that fix linting, various bugs and includes changes about "offering millisecond-timed resolution" (I am curious about what this is about) in the same branch.

unquietwiki:
Will do; I think I might already have something for you. The millisecond
thing: SonarLint was flagging "testable DateTime", so I researched that &
found NodaTime. Compared that to the code, and found there was some
rounding to millisecond stuff happening, so I tried to use the more precise
method. That particular lib goes into ticks & nanoseconds, if you want that
(not sure KSP cares at that depth).

sarbian:
DateTime is used in 4 files :
 - [RCSSolver.cs](https://github.com/MuMech/MechJeb2/blob/0eba6e47b892926d7a57cf8cefe4ce1da00d08ee/MechJeb2/RCSSolver.cs) to see how long the RCSSolver thread ran for (Stopwatch is a better match for this but it is hardly critical).
 -  [MechJebModuleRoverController.cs](https://github.com/MuMech/MechJeb2/blob/fd875e56665927a4723b8c2fab6e6c5796b89f97/MechJeb2/MechJebModuleRoverController.cs) to avoid updating the ETA each frame. (could use the native Unity Time here)
 - [MechJebModuleFlightRecorder.cs](https://github.com/MuMech/MechJeb2/blob/0eba6e47b892926d7a57cf8cefe4ce1da00d08ee/MechJeb2/MechJebModuleFlightRecorder.cs) to get a timestamp for the csv export file name.
 - [MechJebModuleWaypointWindow.cs](https://github.com/MuMech/MechJeb2/blob/0eba6e47b892926d7a57cf8cefe4ce1da00d08ee/MechJeb2/MechJebModuleWaypointWindow.cs) to animate the coordinate gizmo (could use the native Unity Time here)

None of those requires any kind of precision. I would use Time or Stopwatch but the current code is still perfectly valid. 
You need to look at what the code does before opting to add yet an other dependency.

unquietwiki:
@sarbian Cool beans. Well, I'll close this commit & integrate your information on the time stuff into the branch work; as well as split off what I'd like to submit next. Will try to hit you up on Twitter or something. Thanks!

PR #1524, opened 2022-03-19
Better support for Principia node burning and RF spoolup time

NathanKell:
This PR does two things:
* It fixes Principia node burning to actually burn the required delta V (which involves still burning even after Principia nukes the associated node).
* It handles RF engines having spoolup times (by starting burns early). Spool time weighted by thrust is stored in stage stats.


NathanKell:
Forgot I had siimav's stuff in here. Fixed.

PR #1526, opened 2022-03-28
Refresh rate updated when custom window is loaded.

ozraven:
I noticed the new refresh rate wasn't being applied when a custom window was loaded.


DRVeyl:
Looks right to me.  (My bad, I intentionally made that not a property and missed initializing it.)

PR #1542, opened 2022-04-25
Fix much Torquestuff

lamont-granquist:
- convert Vector6 to a struct
- convert availableTorque to a Vector6
- convert Vector6 ToVector3d in a few places to maintain backcompat
  (places that may continue to be buggy I suspect, but out of scope for now)
- in the BetterController select which torque 'sense' to use based on
  the sign of the component of the error vector (asymmetric torque
  handling)
- Abs() the values coming off of GetPotentialTorque() and fix the sense
  of the vector manually to work around KSP bugginess with pos/neg
  values being all over the place.


lamont-granquist:
4 years later, I don't seem to be getting to this at all

PR #1543, opened 2022-04-28
Don't release clamps if an engine failed to ignite

NathanKell:
(And add back the regular execute node buttons when Principia is installed because @lpgagnon misses them.)

Rather than reflecting into TF/TL/whatever, we just simply detect if an engine's stage is >= the current active stage, and if it is not ignited we do not release the next (clamp) stage. This does not interfere with normal clamp decoupling but does prevent decoupling when one or more engines fail to light (due to TF ignition failures or whatever).

This prevents the rocket from falling down and exploding real good, which seems useful.


lamont-granquist:
LGTM

PR #1550, opened 2022-05-13
Start of MechJebLib

lamont-granquist:
Moves the Gooding solver into it and sets up the static methods.

Does some relicensing to GPLv2/GPLv3/MIT since nobody else has touched
these files.


lamont-granquist:
@sarbian you good with the license?  I made it MIT-0 (no attribution) with GPLv2+later for maximum reuseability.

I'll probably pull some MJ code into this directory which can't be relicensed so it won't get the header on top of the code, and may rip a GPLv3 header comment across the rest of the codebase if it becomes confusing.

sarbian:
Yes, I am ok with that license choice.

PR #1551, opened 2022-05-16
Some changes to PVG plane targetting

Nazfib:
- Fix the time-to-plane function to work for all combinations of inclinations and latitudes
- Fix a small bug in PVG's boundary conditions that caused it to launch into the wrong plane when a negative inclination is specified
- Use the sign of the inclination to select between the northern or southern launch window, instead of always going for the nearest of the two (only in PVG launch-to-LAN mode)

One more thing: when warping to a plane, it currently warps until the launch site is exactly on the target plane. However, since the launch vehicle already has some eastwards velocity due to the rotation of the Earth, the rocket is actually east of the plane during most of the launch - this causes PVG to perform a small dog-leg of a degree or so. As far as I know, the ideal time to launch would be approximately half the time-to-orbit _before_ the launch site is in the target plane.

Would it be useful to add an input field in the auto-warp menu, where the user can input an offset that is added/subtracted from the auto-warp time? A default value of 0 gives the behavior as it is now, but by tuning this value, a slightly more efficient ascent is possible.

Another side-effect would be that this makes PVG able to do an Apollo-style off-plane launch to rendez-vous after the moon landing: by varying the offset until the time-to-launch is barely positive (instead of almost a month away), PVG will perform the required dog-leg during ascent automatically (the offset would be approximately the time spent on the surface).

### DIFF
--- a/MechJeb2/MechJebModuleAscentGuidance.cs
+++ b/MechJeb2/MechJebModuleAscentGuidance.cs
@@ -457,7 +457,7 @@ namespace MuMech
                 {
                     launchingToMatchLAN = true;
                     autopilot.StartCountdown(vesselState.time +
-                            SpaceMath.MinimumTimeToPlane(
+                            SpaceMath.TimeToPlane(
                                 mainBody.rotationPeriod,
                                 vesselState.latitude,
                                 vesselState.celestialLongitude,
@@ -474,7 +474,7 @@ namespace MuMech
                     {
                         launchingToLAN = true;
                         autopilot.StartCountdown(vesselState.time +
-                                SpaceMath.MinimumTimeToPlane(
+                                SpaceMath.TimeToPlane(
                                     mainBody.rotationPeriod,
                                     vesselState.latitude,
                                     vesselState.celestialLongitude,
--- a/MechJeb2/MechJebModuleAscentPVG.cs
+++ b/MechJeb2/MechJebModuleAscentPVG.cs
@@ -144,7 +144,8 @@ namespace MuMech
 
             ConvertToSMAEcc(autopilot.desiredOrbitAltitude, DesiredApoapsis, out sma, out ecc);
             ConvertToVTRT(sma, ecc, AttachAltFlag ? attachAlt : autopilot.desiredOrbitAltitude, out gammaT, out rT, out vT);
-            double inclination = autopilot.desiredInclination;
+            // Negative inclination is used in the warp-to-launch code to select the window, but breaks targetting a fixed plane.
+            double inclination = Math.Abs(autopilot.desiredInclination);
 
             if (AscentGuidance.launchingToPlane && core.target.NormalTargetExists)
             {
--- a/MechJeb2/SpaceMath.cs
+++ b/MechJeb2/SpaceMath.cs
@@ -32,23 +32,51 @@ namespace MuMech
         /// </summary>
         ///
         /// <param name="rotationPeriod">Rotation period of the central body (seconds).</param>
-        /// <param name="latitude">Latitude of the launchite (degrees).</param>
-        /// <param name="celestialLongitude">Celestial longitude of the current position of the launchsite.</param>
+        /// <param name="latitude">Latitude of the launch site (degrees).</param>
+        /// <param name="celestialLongitude">Celestial longitude of the current position of the launch site.</param>
         /// <param name="LAN">Longitude of the Ascending Node of the target plane (degrees).</param>
         /// <param name="inc">Inclination of the target plane (degrees).</param>
         ///
         public static double TimeToPlane(double rotationPeriod, double latitude, double celestialLongitude, double LAN, double inc)
         {
-            // alpha is the 90 degree angle between the line of longitude and the equator and omitted
-            double beta = OrbitalManeuverCalculator.HeadingForInclination(inc, latitude) * UtilMath.Deg2Rad;
-            double c = Math.Abs(latitude) * UtilMath.Deg2Rad; // Abs for south hemisphere launch sites
-            // b is how many radians to the west of the launch site that the LAN is (east in south hemisphere)
-            double b = Math.Atan2( 2 * Math.Sin(beta), Math.Cos(beta) / Math.Tan(c/2) + Math.Tan(c/2) * Math.Cos(beta) ); // napier's analogies
-            // LAN if we launched now
-            double LANnow = celestialLongitude - Math.Sign(latitude) * b * UtilMath.Rad2Deg;
+            // Convert latitude and inclination to radians
+            latitude *= UtilMath.Deg2Rad;
+            bool launchSouth = inc < 0;
+            inc = Math.Abs(inc) * UtilMath.Deg2Rad;
 
+            // angleEastOfAN is the difference between the LAN and the (celestial) longitude of the launch site. This value only depends on the launch site latitude
+            // and inclination; the time we seek is exactly when the celestial longitude of the launch site is equal to LAN + angleEastOfAN.
+            double angleEastOfAN;
+            if (inc <= Math.Abs(latitude)) // Inclination is not reachable from this launch site; launch is directly east
+                angleEastOfAN = latitude < 0 ? 1.5 * Math.PI : 0.5 * Math.PI;
+            else if (Math.PI - inc <= Math.Abs(latitude)) // Inclination is not reachable from this launch site; launch is directly west
+                angleEastOfAN = latitude < 0 ? 0.5 * Math.PI : 1.5 * Math.PI;
+            else
+            {
+                // Uses Napier's rules for spherical trigonometry. This is case (R4) from https://en.wikipedia.org/wiki/Spherical_trigonometry#Napier's_rules_for_right_spherical_triangles
+                // where A = inc, a = latitude and b = angleEastOfAN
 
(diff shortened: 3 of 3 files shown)


lamont-granquist:
Cool, I was getting just about ready to tackle this.

Can I get your permission to move this code to MechJebLib/Maths/Functions.cs at some point in the future and have it covered by the 0-license MIT + GPLv2+later licenses?  I'm trying to extract some parts of MJ that I wrote for better reusability outside of MJ.  See https://github.com/MuMech/MechJeb2/pull/1550

Nazfib:
> Can I get your permission to move this code to MechJebLib/Maths/Functions.cs at some point in the future and have it covered by the 0-license MIT + GPLv2+later licenses? I'm trying to extract some parts of MJ that I wrote for better reusability outside of MJ. See #1550

Yes, absolutely, go ahead.

lamont-granquist:
Well actually I'm not really sure why you've removed the MinimumTimeToPlane function and the call to OrbitalManeuverCalculator.HeadingForInclination(inc, latitude) which should handle things like due east / due west already.

There already is support for the launchLANDifference here to lead the launch by a number of degrees:

https://github.com/MuMech/MechJeb2/blob/af7f1dcf274a44b5477a869b94732e89ea138af9/MechJeb2/MechJebModuleAscentGuidance.cs#L448=

It seems like what you want is a launch-immediately option to launch-to-plane, which should then just entirely bypass this whole calculation.

lamont-granquist:
I think I need to write a pile of tests for this function.  I also need to page back into my brain the use of the various napier's rules for spherical trig since I think there was a reason I picked the one I did.

lamont-granquist:
Yeah I wrote this test and its failing on yours:

```
Functions.MinimumTimeToPlane(PERIOD,28.608,0,180,90).ShouldEqual(0, ACC);
```

```
Xunit.Sdk.EqualException
Assert.Equal() Failure
Expected: 0
Actual:   43082.045250000003
```

(ignore the name change, I renamed it MinimumTimeToPlane so I can swap backwards and forwards easily)

Nazfib:
> Well actually I'm not really sure why you've removed the MinimumTimeToPlane function 

I actually mentioned this in the commit message a6db2fcc8d12de2c93e3bedc49a6b32053f48bed: as far as I understand, MechJeb uses the convention that a negative inclination means that the southwards launch window should be taken. Otherwise, there is no way to specify a north or southwards launch.

> and the call to OrbitalManeuverCalculator.HeadingForInclination(inc, latitude) which should handle things like due east / due west already.

I basically re-derived the formula for the angle on paper, and the launch azimuth was not required for it.

> There already is support for the launchLANDifference here to lead the launch by a number of degrees:
>
> https://github.com/MuMech/MechJeb2/blob/af7f1dcf274a44b5477a869b94732e89ea138af9/MechJeb2/MechJebModuleAscentGuidance.cs#L448=

Alright, but then there should also be a way to specify this additional angle when launching to a manually specified LAN. The LAN for the auto-warp should be different from the LAN used by PVG.

> It seems like what you want is a launch-immediately option to launch-to-plane, which should then just entirely bypass this whole calculation.

That would indeed be very useful.

> I think I need to write a pile of tests for this function. 

I recreated the calculation in [GeoGebra](https://www.geogebra.org/calculator/uxzbpdpt), which might at least for the moment give you confidence that the application of Napier's rule is correct. Should I move the function to MechJebLib and add some tests?

> I also need to page back into my brain the use of the various napier's rules for spherical trig since I think there was a reason I picked the one I did.

I don't believe there are any numerical issues with this method (and I explained in the comments why), but if there are any I'd like to know.

> Yeah I wrote this test and its failing on yours:

I don't believe it is wrong - this is the difference between `MinimumTimeToPlane` and 'TimeToPlane`. When launching _north_ to a polar orbit, the LAN you get is equal to your current celestial longitude. To get a LAN of 180° when your current celestial longitude is 0° should require either launching south or waiting half a day.

lamont-granquist:
I'm already writing a bunch of tests for this.  Definitely found one clear quadrant bug in the existing code.

The historical way that launch to plane has worked is that the launch timing has always selected the nearest crossing to minimize the time to launch and ignored the positive/negative inclination and just hit the plane in the first available window.

Nazfib:
> The historical way that launch to plane has worked is that the launch timing has always selected the nearest crossing to minimize the time to launch and ignored the positive/negative inclination and just hit the plane in the first available window

If we want to keep that behaviour, I'll revert the third commit.

lamont-granquist:
> I recreated the calculation in [GeoGebra](https://www.geogebra.org/calculator/uxzbpdpt),

I just get "opening file failed" in that for both firefox in a private window and chrome.

Nazfib:
> > I recreated the calculation in [GeoGebra](https://www.geogebra.org/calculator/uxzbpdpt),
> 
> I just get "opening file failed" in that for both firefox in a private window and chrome.

Sorry, it was still set to 'private'. Can you access it now?

lamont-granquist:
Okay, I think I've managed to show to myself that both versions of `TimeToPlane` are actually equivalent, and I think correct, except for the fix to the planet rotating backwards which was broken previously.

Your version is much simpler spherical trig.  I think I was concerned about the infinities in the tangents and reflexively wanted something based on atan2 to get the quadrants correct.

I think I've just about got a pile of tests and a slightly cleaned up version almost ready to go.

However, there's no bugfix there to explain the issues that people had.  So it is likely that the issue is more related to the other two fixes.  My suspicion is that the initial guess of the heading angle that PVG uses to bootstrap the optimizer is getting northgoing and southgoing confused.  That suggests to me that your fix to allow the user to select northgoing or southgoing smells correct, although that makes me surprised about the other fix to take the Abs() doesn't break it again, but I haven't dug into that code at all yet.

lamont-granquist:
Okay I just fixed TimeToPlane and added a pile of tests.  I think my brain is a little bit fried now though since it took me all day to sort it out.

That doesn't fix any behavior though that would have been causing user bugs, so the other fixes here should get looked at, but sometime after I recharge.

Nazfib:
>Okay I just fixed TimeToPlane and added a pile of tests. I think my brain is a little bit fried now though since it took me all day to sort it out.

Glad I could help a little, sorry that it all got a bit muddled by me putting multiple fixes into the same PR and not properly explaining what I did at first.

> That doesn't fix any behavior though that would have been causing user bugs, so the other fixes here should get looked at, but sometime after I recharge.

Let me see if I can explain what I think the Abs fixes, for when you have the time.

For targetting an orbital plane, PVG calculates the normal vector `[sin(lan) * sin(inc), -cos(lan)*sin(inc), cos(inc)]` ([here](https://github.com/MuMech/MechJeb2/blob/b046641f82e0058f19fd6cb2466d272d13d2bb89/MechJeb2/Pontryagin/PontryaginLaunch.cs#L252) for example) and ensures in the boundary conditions that the final angular momentum vector is aligned with this normal. This requires that the inclination is between 0° and 180°; if it is negative, the ascending and descending nodes are flipped. This causes the LAN to be 180° off from the requested value.

I have run somewhat exhaustive tests today with this fix implemented, from both KSC (northern hemisphere) and Woomera (southern hemisphere):
 1)  -180° < inc < -90° (retrograde orbit, southern window)
 2)  -90° < inc < -|latitude| (prograde orbit, southern window)
 3)  |inc| < |latitude| (prograde orbit, with a dog-leg)
 4)  |latitude| < inc < 90° (prograde orbit, northern window)
 5)  90° < inc < 180° (retrograde orbit, northern window)
 6)  |180° - inc| < |latitude| (retrograde orbit, with a dog-leg)
 7)  inc = 90° (polar orbit, northern window)
 8)  inc = -90° (polar orbit, southern window)
 
In the final case, with inclination close to -90°, PVG failed to find a solution with the error "Unreachable even with infinite Isp" - without the fix, PVG launches to an orbit with the LAN 180° off from the requested value. In all other tests, for all combinations the LAN is correct after reaching orbit. Without the fix, cases (1), (2) and (3) would result in the wrong orbit normal vector.

> My suspicion is that the initial guess of the heading angle that PVG uses to bootstrap the optimizer is getting northgoing and southgoing confused.

This sounds like it might be the cause of the issue for near-polar orbits - I'll have a look tomorrow to see if I can confirm this.

lamont-granquist:
So if you want to update the PR to just have the fix to remove MinimumTimeToPlane from PVG and to implement the launch-now or launch-after-fixed-time that'd probably be fine.   That feels like the correct direction to me.

lamont-granquist:
Oh except maybe open another PR or something since I want to keep around the desiredInclination fix to inspect.

lamont-granquist:
> Would it be useful to add an input field in the auto-warp menu, where the user can input an offset that is added/subtracted from the auto-warp time? A default value of 0 gives the behavior as it is now, but by tuning this value, a slightly more efficient ascent is possible.

I would suggest replacing the existing degree offset with a timed offset, so that "150 s" could be entered.  Breaks the existing behavior but that probably makes way more sense since burntime-to-orbit/2 is probably close enough.

Nazfib:
> So if you want to update the PR to just have the fix to remove MinimumTimeToPlane from PVG and to implement the launch-now or launch-after-fixed-time that'd probably be fine. That feels like the correct direction to me.

#1554 fully fixes launch-to-LAN for all cases, while keeping current behavior.
I'll put the change to keep the sign from the user input in a separate PR.
I'll also create a separate PR for a launch window time offset.

PR #1560, opened 2022-05-30
Fix Launch-into-plane-of-target with Principia

Nazfib:
When launching into the plane of a target with Principia installed, PVG gets stuck on 'INITIALIZING' and never actually launches.

This was previously fixed in cd3d194760f7875658e7e04690e37113d2ca1cd6 by adding the `targetInc` parameter to the PVG constraint setup functions, but the refactor in 136964af2380013e860a04e0433fe6a607874847 broke the fix.


lamont-granquist:
that makes sense.

yet more fodder for the argument this codebase needs more internal tests.

PR #1573, opened 2022-08-31
Fix Principia multiple nodes bug

### LINKED ISSUE #1561: Improve Principia node execution (fix multiple nodes bug)
When using Principia with multiple nodes in the flight plan, sometimes, Mechjeb can screw up the burn.

The Principia timing on the active engine can occasionally be off and so MJ needs to kill rot in case the node was too short. It does that correctly if the node just disappears, i.e. in the case of the last node in the flight plan.

However, in the case of further nodes, principia doesn't delete the node but replaces it with the next one. Then Mechjeb gets confused, cuts the engine, re-aligns itself with the next node and makes a 2nd burn for the rest of the original Δv. This both underburns and wastes  another ignition.

The root cause, I believe is this code snippet: https://github.com/MuMech/MechJeb2/blob/11dee5d6ff22d155c3690561a6644d3764202790/MechJeb2/MechJebModuleNodeExecutor.cs#L162-L174

Specifically, line 162 should have another guard, according to: https://github.com/mockingbirdnest/Principia/issues/2324#issuecomment-609012686

> This isn't a problem whatsoever on kOS. You need to lock steering to the moral equivalent of,
> 
> ```
> function node_steering {
>   if not HASNODE
>     return SHIP:FACING:FOREVECTOR.
>   return NEXTNODE:DELTAV:DIRECTION
> }
> ```
> 
> Once you start the burn,
> 
> ```
> wait until not HASNODE or NEXTNODE:ETA > 1.
> ```
> 
> The second part of this condition corresponds to having more than one maneuver stored.
> 
> This makes the burn script significantly simpler than the equivalent not using Principia.

So, I believe, we shouldn't be assigning `node` to `maneuverNodes[0]` if the ETA is in the future.

rnlahaye:
Fixes #1561.
Thanks to @DanielFabian for finding the root cause.


lamont-granquist:
replaced by #1796

PR #1587, opened 2022-11-14
additional informational items in the target window

### LINKED ISSUE #1586: TLI with Principia
As you surely know, Principia makes Trans-Lunar Injections much more difficult than the stock physics - because of the nodal precession. Moon's inclination varies between 18.134° and 28.725°.

There are basically three options for an efficient TLI from a higher latitude (ie not from the French launch site in Kourou):
* If you launch from Cape Canaveral around the Apollo 11 mission - during 1969's Major Lunar Standstill - the Moon is almost perfectly aligned and nothing is needed except to launch into the right LAN - you can launch every day
* In all other cases one must do what some call a Baikonur-style launch - wait until the Moon is 51° before its AN or DN - then launch into its LAN - add 4 days for a Hohmann transfer - the Moon advances 51° - and you will intercept it right at its AN/DN - in this case you have two launch windows per draconic month
* The Russians launched some of their very early probes from a polar orbit - which allows to launch from any latitude every day, but this is a very hard launch as you have to time your position along the orbit

Normally the second launch should be doable with MechJeb - except I can't find any information window that has what I need - either the time or the angle of the Moon relative to its AN/DN relative to the equator. Am I missing something? I am willing to submit a PR, but I want to make sure that this is not possible before investing time.

mmomtchev:
closes #1586


mmomtchev:
Now everything looks good, but I will try to make it work for other bodies then the Moon - where you need AN/DN relative to the plane of the ecliptic.

Also, it seems that the usual reference point for the Moon is also the plane of the ecliptic, even if in our case the equator is much more useful - but this means that most values that people find on Wikipedia are off.

mmomtchev:
This PR includes the other one because it won't work without it, ping me if you merge the other one before this one, so that I can rebase/merge it

mmomtchev:
in fact it seems that all bodies have equatorial AN/DN and not relative to the ecliptic, but this is KSP
i will probably submit another PR that computes those relative to the ecliptic, this is more than useful in its current state, I am able to execute a perfect TLI with it

lamont-granquist:
yeah you should probably rebase this to remove the other PR

lamont-granquist:
excellent

Issue #1588, opened 2022-11-15
Equatorial DN is off


mmomtchev:
#1587 adds AN/DN to the target orbit info window, which are computed by MechJeb, and it seems that the DN is off.

This image shows Moon's orbit - the target's true longitude is 180° from the AN, but the DN is still 4 days ahead. Also, if you subtract the time to DN from the time to AN, the result should be exactly half the orbital period - and it is not.

![mechjeb-bug](https://user-images.githubusercontent.com/31410344/201891772-4d5f4725-1855-4387-aa65-03cc7db8a977.png)

(This is issue seems to be independent from #1587)

mmomtchev:
Shouldn't all calculations involving orbital use the mean anomaly instead of the true anomaly?

mmomtchev:
I think I have found the problem, what is called true anomaly should probably be called true longitude as AFAIK the result of this calculation is the true longitude and not the true anomaly:

https://github.com/MuMech/MechJeb2/blob/dc618a2aa2222b5ba7ee5b83a149d8a2b72ee71d/MechJeb2/OrbitExtensions.cs#L297

It computes the direction of the equatorial crossing which is the Unity's up direction?

mmomtchev:
Related to Principia dynamically modifying the orbital period, everything is like it should be in MechJeb

mmomtchev:
I found the culprit - there are two overloaded functions called `TrueAnomalyFromVector`: one in KSP, that returns an angle in radians and another one in MechJeb that returns an angle in degrees. There is even a todo for fixing it 😄 

https://github.com/MuMech/MechJeb2/blob/dc618a2aa2222b5ba7ee5b83a149d8a2b72ee71d/MechJeb2/OrbitExtensions.cs#L378

The equatorial DN uses the KSP version.

PR #1589, opened 2022-11-15
TrueAnomalyFromVector returns radians

mmomtchev:
This is a temporary fix that has a correct behavior

The long-term solution will be to transform everything related to the true/mean anomaly to radians, otherwise the slope will always be very slippery

#1588

### DIFF
--- a/MechJeb2/OrbitExtensions.cs
+++ b/MechJeb2/OrbitExtensions.cs
@@ -280,7 +280,7 @@ namespace MuMech
         public static double AscendingNodeTrueAnomaly(this Orbit a, Orbit b)
         {
             Vector3d vectorToAN = Vector3d.Cross(a.SwappedOrbitNormal(), b.SwappedOrbitNormal());
-            return a.TrueAnomalyFromVector(vectorToAN);
+            return a.TrueAnomalyFromVector(vectorToAN) * UtilMath.Rad2Deg;
         }
 
         //Gives the true anomaly (in a's orbit) at which a crosses its descending node
@@ -297,7 +297,7 @@ namespace MuMech
         public static double AscendingNodeEquatorialTrueAnomaly(this Orbit o)
         {
             Vector3d vectorToAN = Vector3d.Cross(o.referenceBody.transform.up, o.SwappedOrbitNormal());
-            return o.TrueAnomalyFromVector(vectorToAN);
+            return o.TrueAnomalyFromVector(vectorToAN) * UtilMath.Rad2Deg;
         }
 
         //Gives the true anomaly at which o crosses the equator going southwards, if o is east-moving,
--- a/MechJeb2/OrbitExtensions.cs
+++ b/MechJeb2/OrbitExtensions.cs
@@ -321,7 +321,7 @@ namespace MuMech
         //angle of the hyperbola.
         public static bool AscendingNodeExists(this Orbit a, Orbit b)
         {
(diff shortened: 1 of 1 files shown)


mmomtchev:
I found other parts that are now broken. Maybe try to change everything to radians? Otherwise it will always be a nightmare.

mmomtchev:
In the second solution I change only the DN calculation (and the misleading comments) - which was purely wrong - this should not have any impact anywhere else. AN was already saying degrees, but returning radians, DN was saying degrees and was adding +180 to an angle in radians.

MechJeb2/OrbitExtensions.cs
```diff
@@ -285,15 +285,15 @@ public static double AscendingNodeTrueAnomaly(this Orbit a, Orbit b)
 
         //Gives the true anomaly (in a's orbit) at which a crosses its descending node
         //with b's orbit.
-        //The returned value is always between 0 and 360.
+        //The returned value is always between 0 and 2 * PI.
         public static double DescendingNodeTrueAnomaly(this Orbit a, Orbit b)
         {
-            return MuUtils.ClampDegrees360(a.AscendingNodeTrueAnomaly(b) + 180);
+            return MuUtils.ClampRadiansTwoPi(a.AscendingNodeTrueAnomaly(b) + Math.PI);
```
lamont-granquist, line 291: This is probably the wrong way to go about doing this.  Most of MJ internally uses degrees instead of Radians, and without going through and auditing all the consumers of these APIs the right way to fix it is probably to return degrees everywhere.  Generally most of the Unity and KSP APIs also return degrees instead of radians.

Radians are objectively mathematically better, but its easier not to fight it.

lamont-granquist: Also are you sure that this returns radians?  It looks like it calls TrueAnomalyFromVector() which looks like it uses degrees.

mmomtchev: There are two `TrueAnomalyFromVector` - one in `OrbitExtension` which takes two arguments, and another one in the KSP DLL which takes only one argument. The KSP one returns radians. This is the KSP one.

Now, this is a huge problem. From what I understand - this is the first time I am touching anything inside a KSP mod - KSP used to have degrees everywhere but later switched to radians for the true and mean anomaly - variables that are never visible to the user in the stock game.

There is a TODO that I linked in the issue about switching true/mean anomaly to radians to align to the game.

Anyway, currently those functions work in radians, even if the comment says degrees. And the DN was way off, now it is ok. The DN was not displayed before my other PR - which explains why it went unnoticed.

In order to minimize the impact, I decided to keep those functions in radians and modify only the comment and the DN which was simply wrong.

lamont-granquist: Ah, the same-named functions tripped me up.

lamont-granquist:
Okay so it looks like the four "*Exists" APIs are all returning bools and compare against the KSP API which returns rads now.  It is interesting that `ClampDegrees180` calls in those have been broken for ages, but apparently they must not do anything.  

Then it looks like DescendingNodeEquatorialTrueAnomaly, AscendingNodeEquatorialTrueAnomaly, DescendingNodeTrueAnomaly and  AscendingNodeTrueAnomaly are private APIs that are not used anywhere so those can certainly be changed.

Okay so it looks like this should be safe +/- actually fixing real bugs.

PR #1591, opened 2022-11-27
Fix automatically setting config to new skin on load

SmoothMcDouglette:
Remove line resetting old skin flag on config load.

Was trying to set the old skin but it would keep resetting it in the config file. I found this line, and it seems like it might be the culprit. Not sure if this was put here intentionally or perhaps I have no idea what I'm talking about.

### DIFF
--- a/MechJeb2/MechJebModuleSettings.cs
+++ b/MechJeb2/MechJebModuleSettings.cs
@@ -46,7 +46,6 @@ namespace MuMech
             if (useOldSkin)
             {
                 skinId = 1;
-                useOldSkin = false;
             }
         }
 


lamont-granquist:
Did it fix anything?  Because it doesn't look like that fixes anything.  The "useOldSkin" flag looks like backcompat for old config and all it does is if it is set it flips the skinId to 1 (which I'm guessing is the old skin) and then resets the flag.  If you don't reset the flag there, then if it ever gets set (although it is probably ancientsauce enough that it never triggers) then it would be impossible to unset it and get off of the old skin.  I don't think you've found the right thing.

SmoothMcDouglette:
Yes, sorry, you are right. I misread that. But I'm confused as to how the block starting at line 66 is supposed to be working then:

`GUILayout.Label(Localizer.Format("#MechJeb_Settings_label1", (GuiUtils.SkinType)skinId));//"Current skin: <<1>>"
            if (GuiUtils.skin == null || skinId != 1)
            {
                if (GUILayout.Button(Localizer.Format("#MechJeb_Settings_button2")))//"Use MechJeb 1 GUI skin"
                {
                    GuiUtils.LoadSkin(GuiUtils.SkinType.MechJeb1);
                    skinId = 1;
                }
            }`
            
So, the skin id will only be set to 1 if it's not set to 1 initially?

Issue #1609, opened 2023-03-25
ArgumentException: OrbitExtensions.GetEccentricAnomalyAtTrueAnomaly


mihemihe:
* Using MechJeb2 in combination with a RSS/RO/RP1 Express setup, on  clean 1.12.3.
* Been using this setup for several weeks without issues
* During the past 2 days most of the times I leave the earth SOI and start orbiting the sun I get this exception spammed, to the point the game becomes unresponsive.

This is the issue:

[EXC 16:27:45.409] ArgumentException: OrbitExtensions.GetEccentricAnomalyAtTrueAnomaly: True anomaly of 2.69600474693162 radians is not attained by orbit with eccentricity 2.19997327302083
	MuMech.OrbitExtensions.GetEccentricAnomalyAtTrueAnomaly (Orbit o, System.Double trueAnomaly) (at <24b14848e53c4c4492718006da41870f>:0)
	MuMech.OrbitExtensions.TimeOfTrueAnomaly (Orbit o, System.Double trueAnomaly, System.Double UT) (at <24b14848e53c4c4492718006da41870f>:0)
	MuMech.OrbitExtensions.TimeOfAscendingNodeEquatorial (Orbit o, System.Double UT) (at <24b14848e53c4c4492718006da41870f>:0)
	MuMech.MechJebModuleInfoItems.TimeToEquatorialAscendingNode () (at <24b14848e53c4c4492718006da41870f>:0)
	(wrapper dynamic-method) System.Object.lambda_method(System.Runtime.CompilerServices.Closure,object)
	MuMech.ValueInfoItem+<>c__DisplayClass14_0.<.ctor>b__0 () (at <24b14848e53c4c4492718006da41870f>:0)
	MuMech.ValueInfoItem.UpdateItemCache () (at <24b14848e53c4c4492718006da41870f>:0)
	MuMech.ValueInfoItem.UpdateItem () (at <24b14848e53c4c4492718006da41870f>:0)
	MuMech.MechJebModuleCustomInfoWindow.UpdateWindowItems () (at <24b14848e53c4c4492718006da41870f>:0)
	MuMech.MechJebModuleCustomInfoWindow.WindowGUI (System.Int32 windowID) (at <24b14848e53c4c4492718006da41870f>:0)
	MuMech.DisplayModule.ProfiledWindowGUI (System.Int32 windowID) (at <24b14848e53c4c4492718006da41870f>:0)
[...]

PR #1610, opened 2023-03-27
Fix more Deg/Rad issues in OrbitExtensions

sarbian:
It seems we missed a few and I guess they have been breaking some module for a long time.

Merge to master so I can push a release with those if this looks good.

### DIFF
--- a/MechJeb2/OrbitExtensions.cs
+++ b/MechJeb2/OrbitExtensions.cs
@@ -266,7 +266,7 @@ namespace MuMech
         {
             if (o.eccentricity < 1)
             {
-                return o.TimeOfTrueAnomaly(180, UT);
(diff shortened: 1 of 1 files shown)


lamont-granquist:
fixes #1609?

lamont-granquist:
and related to #1589?

sarbian:
Yes, it most likely fixes #1609. And this complete #1589 by fixing the remaining Deg uses. However I am surprised nothing broke before that PR...

sarbian:
The fix is now in dev and master. I will do a new release, and hopefully I did not break more than I fixed.

lamont-granquist:
I'll take a look at it today, I got woke up yesterday by hammering and construction at 8am.  Need a bit of caffiene first though.

PR #1612, opened 2023-03-30
Fix FuelFlowSim for NO_FLOW resources in Engines

lamont-granquist:
This fixes a bug affecting Delta IV Heavy setup with asparagus staging.  The RS68 engines have had Ablator added to them as an additional "fuel" which is NO_FLOW.

The old check CanDrawResource() which was used to determine if staging was allowed was not aware of fuel modes and that caused it to check if any part in the crossfeedset had the fuel even for NO_FLOW fuels.  This meant that after the side boosters flamed out that the center core would still be burning Ablator and it would find that Ablator "accessible" in its crossfeed set coming from the side engines (because of asparagus).

This may also affect SRBs since those fuels are NO_FLOW as well?

However, I think it requires something like Asparagus staging And given that RO doesn't allow crossfeed through decouplers or fuel pipes it may be a pretty narrow bugfix.

### DIFF
--- a/MechJeb2/FuelFlowSimulation.cs
+++ b/MechJeb2/FuelFlowSimulation.cs
@@ -263,15 +263,22 @@ namespace MuMech
                             return false;
                         }
 
-                        if (n.ContainsResources(burnedResources.value))
+                        foreach (int id in burnedResources.value)
                         {
-                            for (int j = 0; j < activeEnginesCount.value.Count; j++)
+                            if (!n.ContainsResource(id))
+                                continue;
+
+                            for (int j = 0; j < activeEngines.value.Count; j++)
                             {
                                 FuelNode engine = activeEngines.value[j];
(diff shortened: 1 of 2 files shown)


MechJeb2/FuelFlowSimulation.cs
```diff
@@ -263,15 +263,22 @@ private bool AllowedToStage()
                             return false;
                         }
 
-                        if (n.ContainsResources(burnedResources.value))
+                        foreach (int id in burnedResources.value)
```
lamont-granquist, line 266: this had to be changed to loop over each resource, uniquely checking if an activeengine could draw that resource from the part inspected to see if it could be dropped.

lamont-granquist:
Looks like this works and I don't think I've left any debugging in it.  Still have a StagingController bug to track down.

PR #1641, opened 2023-04-20
PID tuning

### LINKED ISSUE #1636: Smart A.S.S. momentum issue
When I tell Smart ass to go to Prograde or any of the nodes, it will completely overshoot the node and is simply not as accurate as it used to be.

lamont-granquist:
- revert the likely somewhat poor idea I had for adding an integral term to the position PID

- refactor the position controller to more clearly allow setting Kp directly rather than whatever `PosFactor` was.

- bump up position Kp from around 0.7 to 2.0 which helps with rockets at physwarp going through max-Q

- add deadband and clegg crossing logic to the PID controller and use that to add some stability to the attitude controller around zero.

closes #1636 maybe, IDK there's not enough info there.


lamont-granquist:
Also removed some warp scaling.  The scaling of N by the warpFactor was just wrong.  The scaling of the warp factor was also likely wrong (probably caused by the previous).

Also removed all the phase angle default stuff, I've never seen much difference using any of the variant settings.

PR #1679, opened 2023-05-27
Update MechJebLib copyright headers

lamont-granquist:
This utilizes the SPDX-License-Identifier to be more permissive and legally clear and extend the license under MechJebLib.

It follows https://cr.yp.to/publicdomain.html in order to place the work under the public domain.

The other licenses are also included to satisfy license checkers and to provide a deliberate fallback for any concerns over the public domain declaration.  Worst case this should be equivalent to dual licensing under 0BSD and LGPL2.1+.

The License Identifier also clears up my poor wording using "and" under the dual license when the intent was to make it at the choice of the user, not to have both always applied (which doesn't make any sense to me which is why I didn't think it needed any clarification)

This work is hereby placed into the public domain.


lamont-granquist:
@sarbian pretty sure you're the only one that has patched any of this still (the Gooding Lambert patches from a long time ago) so you good with the relicensing to be more permissive here?

(And I should probably note for posterity that by using "and the MechJeb contributors" here I'm just creating boilerplate for the future, AFAIK there's still no external contributors under this directory, but I assume there will be in the future)

sarbian:
I am totally fine with it. I never was a fan of the original license.

lamont-granquist:
cool.  should also probably extract MJLib itself at some point to a separate project/dll to keep things nice and clear (plus MJLib doesn't link against Unity at all which allows the tests to run without it)

PR #1687, opened 2023-06-03
Fix autostaging for asparagus boosters

### LINKED ISSUE #1658: Autostaging issue
KSP 1.12.3, MJ#1197

Autostaging doesn't work properly anymore on several time tested rockets: radial attached LfOx booster stages with tanks and engines and fuel lines to the core stage are not separated anymore when their tanks run empty. Could reproduce this on different rockets, e.g.: https://kerbalx.com/VoidSquid/Early-Career-Mun-Lander-1P

Did some tests, last version it worked was 1155, since 1156 all version show this bug.

"Makes me wonder what I broke..." @lamont-granquist

Nazfib:
In this loop, "continue" means "this condition doesn't prevent staging (though another might)"; "return true" means "this tank still contains resources that some active engine needs; don't stage it away yet".

Given that interpretation, the condition on this line is reversed: when there's no more of a resource that can be pulled from this tank, then it is not a problem if we stage the tank away; if there's some of the resource in the tank, then we need to check if any active engine can reach it.

In PR #1613 the other checks in the loop were inverted, but this one was missed.

Fixes #1658.

### DIFF
--- a/MechJeb2/MechJebModuleStagingController.cs
+++ b/MechJeb2/MechJebModuleStagingController.cs
@@ -502,7 +502,7 @@ namespace MuMech
                             continue;
                         PartResource r = p.Resources.Get(propellant.id);
 
-                        if (r.amount > p.resourceRequestRemainingThreshold)
+                        if (r.amount <= p.resourceRequestRemainingThreshold)
                             continue;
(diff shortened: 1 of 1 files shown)


lamont-granquist:
Oh nice that would explain why the effect is somewhat random as well, since it would depend on floating point roundoff errors.

PR #1747, opened 2023-08-14
Tweak DecouplingAnalyzer algorithm

lamont-granquist:
As we walk through the vessel we should never decrease the decoupledInStage value to a lower value based on the inverseStage of the decoupler.

If that happens then what we've done is fire a decoupler that is already decoupled.  In that case the decoupler is rendered "sterile" and is treated like just another part.

This is a departure from old behavior, but is necessary to get some analysis to be correct, and to agree with the autostaging behavior of mechjeb (and to make PVG work correctly).

### DIFF
--- a/MechJeb2/MechJebLib/Simulations/DecouplingAnalyzer.cs
+++ b/MechJeb2/MechJebLib/Simulations/DecouplingAnalyzer.cs
@@ -2,6 +2,7 @@
 
 using System.Collections.Generic;
 using MechJebLib.Simulations.PartModules;
+using static MechJebLib.Utils.Statics;
 
 namespace MechJebLib.Simulations
 {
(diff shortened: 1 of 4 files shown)


MechJeb2/MechJebLib/Simulations/DecouplingAnalyzer.cs
```diff
@@ -70,80 +68,81 @@ private static int CalculateDecoupledInStage(SimVessel v, SimPart p, SimPart? pa
             if (p.DecoupledInStage != int.MinValue)
                 return p.DecoupledInStage;
 
-            for (int i = 0; i < p.Modules.Count; i++)
+            if (p.InverseStage >= parentDecoupledInStage)
```
lamont-granquist, line 71: this is the important line in this PR which was added

lamont-granquist: I wasn't sure about ">=" vs. ">" but I think it doesn't really matter in the '==' case.  Maybe going with ">" would have avoided some useless work though.

PR #1755, opened 2023-08-23
Convert ReturnFromMoon maneuver to use analytical Jacobian

lamont-granquist:
This doesn't seem to do that much in terms of speedup or stability really (although I didn't try investigating stability very well).

Investigating the values in the jacobian one of the biggest problems is clearly integrating back from the terminal endpoints.  The biggest numbers come from the state transition matrix there.  This is likely because small target periapsis constraints results in highly eccentric return orbits.  Taking problems that blow up and using a higher target periapsis often results in convergence.

There is a note in Ellison & Englander (2019) that might be applicable:

> A more natural propagation strategy would be to utilize time
> eegularization, such a Sundman transformation,18–20 and a
> corresponding modification to the variational equations. This
> is left as future work.

It might also be possible to use some kind of "homotopy" and if the problem fails, relax the periapsis constraint to some higher intermediate point that is easier to solve, then use that as an initial guess.  I'd really like to avoid iterative approaches.

Also I have not investigated how good/bad my initial guessing is, and it may be possible that in the cases which do not converge that it is very poor.

I think it may also work to target an Apoapsis now (higher orbit than the moon around the primary), although I didn't test that a lot.


lamont-granquist:
Yeah my initial guess generation is hot, hot garbage.

I'm always using a posigrade hyperbolic vinf target in the Moon SOI and always using a posigrade return orbit and always in the plane of the Moon.  My initial guesses are probably 180 degrees away from the result in some cases.  Should probably try retrograde orbits and should probably study the inc of the return orbits and should consider something like Max(moon_inc, parking_inc) or something like that.

Could also consider doing a smaller SQP problem to find a feasible solution that uses vinf as the optimizer variables and uses the analytic hyperbolic orbit ejection generator to produce a ∆v and ∆t and then propagate that analytically to the SOI and convert to planet-centric coordinates and then compute the PeR and apply a constraint.  Could then minimize J=∆v and come up with a decent feasible guess (which honestly might be a faster and smaller and good-enough alternative).

PR #1771, opened 2023-09-24
Do not chase the maneuver node if there's less than a second of burn left

siimav:


MechJeb2/MechJebModuleAttitudeController.cs
```diff
@@ -315,6 +315,30 @@ public double attitudeAngleFromTarget()
...
+            fwd = Vessel.patchedConicSolver.maneuverNodes[0].GetBurnVector(Orbit);
+            up = Vector3d.Cross(fwd, VesselState.normalPlus);
+            Vector3.OrthoNormalize(ref fwd, ref up);
+            rotRef = Quaternion.LookRotation(fwd, up);
+            rotRef = Quaternion.FromToRotation(thrustForward, VesselState.forward) * rotRef;
+            //up = attitudeWorldToReference(attitudeReferenceToWorld(attitudeTarget * Vector3d.up, reference), reference);
+            //var attitudeTarget = Vector3d.forward;
+
+            return Math.Abs(Vector3d.Angle(rotRef * Vector3d.forward, VesselState.forward));
+        }
```
lamont-granquist, line 340: I'm confused as to why we need this instead of just directly using `attitudeAngleFromTarget()`?

siimav: Because Target becomes whatever Kill Rot wants it to be instead of the actual maneuver node. Remember that the mode is switched if there's less than 1s of burn left.

lamont-granquist:
thought replaces would close this automatically, weird...

PR #1788, opened 2023-10-24
Behavioral changes to Node Executor

lamont-granquist:
The stock node executor no longer "hunts" for the last dribbling bit of the maneuver node and the "tolerance" setting has been eliminated.  There is a 1 second terminal guidance period where tracking is frozen and the burn terminates when the angle between the heading and the maneuver node is >= 90 degrees.

This also backs out the MANEUVER_COT and thrustForward stuff for asymmetric thrust vehicles.  This is because thrustForward isn't valid if the engines are off.  Don't know if VesselState needs fixing to fix thrustForward or if the node executor needs to switch modes based on thrust being on or not, but it feels like VesselState should be updated.

replaces #1771

### DIFF
--- a/MechJeb2/MechJebModuleNodeExecutor.cs
+++ b/MechJeb2/MechJebModuleNodeExecutor.cs
@@ -1,7 +1,8 @@
-﻿using System;
-using JetBrains.Annotations;
+﻿using JetBrains.Annotations;
 using MechJebLib.Simulations;
 using UnityEngine;
+using static System.Math;
+using static MechJebLib.Statics;
 
 namespace MuMech
 {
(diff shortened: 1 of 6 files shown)


MechJeb2/MechJebModuleNodeExecutor.cs
```diff
@@ -223,89 +199,95 @@ public override void OnFixedUpdate()
...
-                        : Core.Attitude.attitudeAngleFromTarget() < 10))
-                {
-                    Core.Warp.WarpToUT(VesselState.time + timeToBurn);
-                }
-                else
-                {
-                    //realign
-                    Core.Warp.MinimumWarp();
-                }
+                Core.Warp.WarpToUT(ignitionUT - 600);
```
NathanKell, line 271: I don't think this is correct; I think you want to check attitude still. In this case we'll gradually step down warp as we get close to 600s from node, then (if we started the warp aligned) step it back up again once we pass < 600s from node (!).

You want to warp all the way until the lead time if you're already aligned; you only want to break at ~600s if you're not aligned already.

PR #1803, opened 2023-10-31
Add SerializeField to private Persistent fields

lamont-granquist:
This might fix the problem with window positions not updating.


lamont-granquist:
Narrator:  It did not.

PR #1846, opened 2023-12-11
Fix custom window presets

Nazfib:
Some fields got renamed since these were last updated; some of these strings were simply wrong (periods that should be colons) and would never have worked.


lamont-granquist:
oh yeah those strings are kinda terrible, that should all really get turned into code and i need to be more careful about that.

Issue #1860, opened 2024-02-10
RCS keeps getting enabled when toggling autopilot and when staging


Gouvernathor:
I didn't find any option to disable that in the ascent autopilot settings.
This is using my RCS reserves without asking for permission, kinda rude.

When I'm on the landing pad, the mod doesn't even allow me to deactivate RCS !

Gouvernathor:
It also triggers during landing guidance, whenever the thrust is very low, it seems. I have to spend a lot of time just shutting off the RCS, it's very annoying and borders unplayable.

Sinomen:
RCS is always active when enabling MJ, be it Ascent Guidance, be it Node Execution. No Kopernicus/RO/RSS installed. This bug is persistent since build #1344.

Starwaster:
Sounds like RCS auto mode is on. There is a setting for that in the attitude adjustment window.

Gouvernathor:
The "RCS auto mode" is disabled in the attitude adjustment window, which I never opened before.
The one in landing guidance is still enabled, but I suppose (as it did before) that that setting only changes anything for specific moments during the landing autopilot, and not to unrelated modules (ascent guidance or auto maneuver for example).

Sinomen:
> Sounds like RCS auto mode is on. There is a setting for that in the attitude adjustment window.
> 
> (and the landing guidance window has a similar RCS control for killing lateral motion)

RCS auto mode is off, never was on. As I said, the issue came up with #1344, never before, and is persistent since.

Starwaster:
@Sinomen Narrow that down to Real Fuels which you can have installed independently of RO or RSS. That is what is being looked for.
That said, it should only ever trigger if commanded throttle is > 0 and there is unstable ullage, so it's clearly missing some checks. There's only 4 files where AutoRCSUllaging is referenced so it shouldn't be too hard to find.

Sinomen:
No Real Fuels installed here.

Starwaster:
If you don't even have RF installed then that is definitely a problem, but it should make it easier to track down. I'm on my phone so I'll have to wait until I'm back on my PC to comment further.

Gouvernathor:
No Real Fuels either on my side.

Starwaster:
https://github.com/MuMech/MechJeb2/blob/12368452ea88c8584f80866bf0ffc55276d4a053/MechJeb2/MechJebModuleStagingController.cs#L300
[...]

PR #1873, opened 2024-03-21
Update VesselState.cs

Starwaster:
There's a bug reported #1860  where RCS keeps turning on when unwanted for non-Real Fuels users and requires turning off AutoRCSUllaging. This may be enough to fix it but this could use other sanity checks elsewhere. Also, RCS ullaging should either

* Check for Core.Attitude.RCS_auto = true or else don't enable RCS!
* Put in code to turn RCS back off when ullaging is completed.

### DIFF
--- a/MechJeb2/VesselState.cs
+++ b/MechJeb2/VesselState.cs
@@ -1384,7 +1384,8 @@ namespace MuMech
             public readonly Vector6 torqueDiffThrottle = new Vector6();
 
             // lowestUllage is always VeryStable without RealFuels installed
-            public double lowestUllage;
+            // Don't wait for Update, set it now so we don't invite a race condition.
+            public double lowestUllage = 1.0;
 
             public struct FuelRequirement
             {


lamont-granquist:
sorry been busy and i didn't check to see how easy this one was.

lamont-granquist:
I think i'd be okay with checking RCS_auto but does it persist?

And for RF it'd be nice to force it to true rather than false somehow.

Starwaster:
> I think i'd be okay with checking RCS_auto but does it persist?
> 
> And for RF it'd be nice to force it to true rather than false somehow.

Pretty sure it does not persist. Not last time I checked and I don't think that ever changed.

lamont-granquist:
Yeah for RO/RF that should really persist, probably should persist anyway.

Because really in RO/RF with PVG coasts I want to make it pretty difficult to disable RCS, rather than having it easy to forget to enable it all the time.

PR #1892, opened 2024-06-04
Fix ToSI for floats

Nazfib:
Fix `float.ToSI` to match the changes to `double.ToSI` in 02fab0d671985f54d3a22d769affa904db3294b6

The float version is only used at
https://github.com/MuMech/MechJeb2/blob/6150a7869e634d8307343fc32c0a1e10ebf21c2e/MechJeb2/MechJebModuleInfoItems.cs#L436
https://github.com/MuMech/MechJeb2/blob/6150a7869e634d8307343fc32c0a1e10ebf21c2e/MechJeb2/MechJebModuleTranslatron.cs#L125
https://github.com/MuMech/MechJeb2/blob/6150a7869e634d8307343fc32c0a1e10ebf21c2e/MechJeb2/MechJebModuleWaypointWindow.cs#L183

All three are relatively obscure, which is probably why it hasn't been noticed before.

### DIFF
--- a/MechJeb2/MechJebModuleWaypointWindow.cs
+++ b/MechJeb2/MechJebModuleWaypointWindow.cs
@@ -180,7 +180,7 @@ namespace MuMech
                 }
             }
 
-            _stats = $"{Count} waypoints over {distance.ToSI(-1)}m";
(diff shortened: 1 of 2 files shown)


lamont-granquist:
yep, probably missed those in the noise.

PR #1922, opened 2024-11-25
Fix PVG settings stage numbers

### LINKED ISSUE #1920: Visual Bug with Earlyshutoff / last stage
KSP_x64_2024-11-22_09-50-10
KSP_x64_2024-11-22_09-50-25

[R-7 Super.zip](https://github.com/user-attachments/files/17868661/R-7.Super.zip)

Here is 2 screenshots and a craft file from RP-1 showing it

Nazfib:
When multiple stages in the Δv stats window have the same KSP stage number, the (unguided) and (optimize) labels get applied to the wrong stage(s). This commit changes the window to apply the labels to the stage with the correct KSP stage, which may not be the same as the index in the vacStats list.

Fix #1920.

### DIFF
--- a/MechJeb2/MechJebModuleAscentPVGSettingsMenu.cs
+++ b/MechJeb2/MechJebModuleAscentPVGSettingsMenu.cs
@@ -51,11 +51,13 @@ namespace MuMech
             {
                 GUILayout.BeginVertical(GUI.skin.box);
 
-                _ascentSettings.LastStage.Val = Clamp(_ascentSettings.LastStage.Val, 0, Core.StageStats.VacStats.Count - 1);
+                _ascentSettings.LastStage.Val = Clamp(_ascentSettings.LastStage.Val, 0, vacStats[vacStats.Count - 1].KSPStage);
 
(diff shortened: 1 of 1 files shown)


MechJeb2/MechJebModuleAscentPVGSettingsMenu.cs
```diff
@@ -64,9 +66,9 @@ protected override void WindowGUI(int windowID)
 
                     GUILayout.BeginHorizontal();
                     GUILayout.Label($"{stats.KSPStage,3} {stats.DeltaV:##,###0} m/s");
-                    if (_ascentSettings.UnguidedStages.Contains(mjPhase))
+                    if (_ascentSettings.UnguidedStages.Contains(stats.KSPStage))
                         GUILayout.Label(" (unguided)");
-                    if (_ascentSettings.OptimizeStage == mjPhase)
+                    if (_ascentSettings.OptimizeStage == stats.KSPStage)
```
lamont-granquist, line 71: this line might cause a problem if it matches multiple stages.  i don't know that it is fixable here, though.

really the fix that needs to happen is that the FFS needs the asymmetric burnout code added to it, in order to suppress the extra stages.

PR #1938, opened 2025-02-28
Coast fixes

Nazfib:
When the current stage (in case of 'coast before') or the next stage (in
case of 'coast after') occurs multiple times in the stage stats, PVG
inserts a coast before/after _each_ of them. This should not happen,
there should only ever be one coast stage.

When a coast is inserted immediately after the "Early shutoff stage",
RCS does not get enabled and the `StartCoast` variable doesn't get set;
this means that the coast time does not tick down.

This patch fixes both these issues.


lamont-granquist:
👍

PR #1939, opened 2025-03-02
Fix saving and loading of vessel-specific settings

Nazfib:
When loading a craft file in the Editor, there's no vessel name available; thus, all vessel-specific settings (those with persistent pass TYPE) get loaded at their default values. A few seconds later, those get saved (with the ship name at the top of the screen), overwriting any previously vessel-specific settings with their default values.

Since there's no way to get a ship name in the OnLoad method (it gets called before the Editor has set the name from the .craft file), the best option is to simply not save the settings from inside the editor at all.

### DIFF
--- a/MechJeb2/MechJebCore.cs
+++ b/MechJeb2/MechJebCore.cs
@@ -959,13 +959,15 @@ namespace MuMech
                 Profiler.BeginSample("MechJebCore.OnSave.sfsNode");
                 if (sfsNode != null) sfsNode.nodes.Add(local);
                 Profiler.EndSample();
+
+                // Vessel-specific settings don't get saved in the Editor.
+                // The reason for this is, that OnLoad has no way to determine the vessel name in the Editor; this means
+                // that the vessel settings get loaded at their default values, and then shortly afterward saved -
+                // overwriting the previously saved settings!
                 Profiler.BeginSample("MechJebCore.OnSave.type");
-                // The EDITOR => FLIGHT transition is annoying to handle. OnDestroy is called when HighLogic.LoadedSceneIsEditor is already false
-                // So we don't save in that case, which is not that bad since nearly nothing use vessel settings in the editor.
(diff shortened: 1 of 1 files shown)


lamont-granquist:
Is there any way to fix that right so that the loading of settings is delayed or not done if the vessel name isn't valid?

Nazfib:
That is basically what it did: there is no valid vessel name, so it just loaded the default values for all the settings. The problem is that those default settings then get saved (using the text field at the top of the window as the vessel name), overwriting whatever custom settings were previously there.

Unfortunately, during the OnLoad call, there's no vessel name available; the field that is accessed during OnSave still has the name of the previous ship loaded in the VAB at this point (when loading a craft file), or null (when doing a "Revert to VAB") (IIRC, it may have been the empty string). There's no way to detect whether this is a "valid" name, or whether this name is supposed to be for a different vessel. Falling back to not loading anything is what causes this bug in the first place.

To fix this such that vessel-specific settings get saved and loaded in the Editor as well, another call to OnLoad would need to be done at a later time, after loading the craft file. I'm not exactly sure what the order between this and OnStart is in various different circumstances (creating a new part, loading an existing craft file from the menu, reverting from the flight scene, entering the VAB from the space center but with a craft already there) — there's already some special handling there. There then needs to be a way to make sure that OnSave does not save the settings before OnLoad has run for the second time.

I couldn't find any craft-specific settings that would have any effect in the VAB. This problem has existed for a very long time, and it has never (as far as I can tell) been noticed, until PVG got some settings that are saved _only_ per-type, not per-part or globally.

lamont-granquist:
This is weird because it also seems to undo all the windows and everything.  I swear it didn't do this before...

Anyway, now I get it or at least see how it's super annoying and something needs to get done (and the problem is that to do it perfectly right we'd really need to split settings across and VAB-vs-flight and save only the VAB settings in the VAB, which sounds like ugh to me).

PR #1990, opened 2025-08-07
Some more stage numbering fixes

Nazfib:
Similar to #1922


lamont-granquist:
yeah got it, comes from breaking the 1:1 relationship between mj stages and ksp stages

PR #1991, opened 2025-08-10
Update bettercontroller tuning

lamont-granquist:
- velocity pid now outputs angular accel instead of actuation
- removed buggy compensation for higher warp [KSPs own crazy tweaking of gimbal rate seems to compensate fairly well for higher warp]
- tuned bandwidth of the PIDs down significantly (4 rad/sec for inner, 1 rad/sec for outer) with reference-tracking focus and highest phase margins [which produces minimal overshoot for snappy RCS, while keeping the PIDs from being too snappy for slow gimbals]
- design was done against a no-actuator-delay/no-gimbal system (RCS/RW), with the slow bandwith being the tuning for possible delay.

### DIFF
--- a/MechJeb2/AttitudeControllers/BetterController.cs
+++ b/MechJeb2/AttitudeControllers/BetterController.cs
@@ -10,10 +10,10 @@ namespace MuMech.AttitudeControllers
 {
     internal class BetterController : BaseAttitudeController
     {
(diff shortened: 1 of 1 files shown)


lamont-granquist:
For future reference, with a gimbal rate of 8, the filter coefficient is 0.16 and the bandwidth of the actuator is 8 rad/sec.

PR #1996, opened 2025-08-13
Landing guidance improvements

njits23:
Currently, landing guidance loves wasting ignitions on limited ignition engines. It also doesn't wait for a craft to stop rotating before entering timewarp.

This PR adds a mode to limit the minimum throttle to prevent a full engine shutdown whilst the landing autopilot is performing burns. It will now only ignite three times in all cases I've tested.

If you turn off the minimum throttle limit, the old throttling behaviour is restored.

This PR also 
1. Makes MechJeb wait for the craft's rotation to settle before warping.
2. Moves the braking end altitude to 200 meters. 500 meters was excessive (open to reverting this based on a good argument).
3. Removes the warp between the horizontal velocity kill and final descent, which always resulted in the craft crashing, and isn't necessary for a few hundred meter final descent.

### DIFF
--- a/MechJeb2/LandingAutopilot/DecelerationBurn.cs
+++ b/MechJeb2/LandingAutopilot/DecelerationBurn.cs
@@ -48,7 +48,7 @@ namespace MuMech
                     return this;
                 }
 
-                if(!_decelerationBurnTriggered)
+                if (!_decelerationBurnTriggered)
                     _decelerationBurnTriggered = true;
 
                 Vector3d desiredThrustVector = -VesselState.surfaceVelocity.normalized;
@@ -61,7 +61,7 @@ namespace MuMech
                 if (Vector3d.Dot(VesselState.surfaceVelocity, VesselState.up) > 0
                     || Vector3d.Dot(VesselState.forward, desiredThrustVector) < 0.75)
                 {
-                    Core.Thrust.TargetThrottle = 0.01F;
+                    Core.Thrust.TargetThrottle = Core.Landing.MinAllowedThrottle();
                     Status                     = Localizer.Format("#MechJeb_LandingGuidance_Status5"); //"Braking"
                 }
                 else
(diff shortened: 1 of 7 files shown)


lamont-granquist:
i think that is there so that users can override and kill the throttle manually.  it should still work, although it may require that the default throttle settings in KSP are set to >0%

njits23:
If that's the intention, as far as I can tell it doesn't work that way because that code block sets the s.mainThrottle value to the TargetThrottle earlier in the file:
https://github.com/MuMech/MechJeb2/blob/d79924f5f7d6ad7b866139f5b5bcc7f738ae2c66/MechJeb2/MechJebModuleThrustController.cs#L448

lamont-granquist:
Well that's @#$%^ wild...  I could swear I tested the min throttle setting before and it worked, but i certainly don't see how it could...

lamont-granquist:
(and I went back like 8 years and doesn't look like it worked at all before e.g. ullage support was added, etc)

lamont-granquist:
Ohhh right, so minthrottle is the minimum throttle that a throttle limiter can set.  So if you've requested 100% throttle, and the g-limiter wants to cut it down to 1%, and you've got it set to 5% then it'll limit the limiter to 5%.  It never filtered the requested non-zero throttle to keep it above zero though.

njits23:
Right. Does that mean you still want changes to my implementation of the minimum throttle in the lander autopilot, or is it good enough? I didn't want to touch the main throttle controller in case making the limit work the way I needed it to for the landing autopilot broke other behaviour. I wanted to keep the changes contained to landing autopilot files only.

lamont-granquist:
I'm thinking the ThrustController should really have an API added, something like:

```
void RequestThrottle(float target, bool enforceMinimum = true, bool allowZero = false) {
  if (enforceMinimum && LimiterMinThrottle) {
    if (allowZero && target == 0) {
      TargetThrottle = 0;
    } else {
      TargetThrottle = Mathf.Max(request, MinThrottle);
  } else {
    TargetThrottle = target;
  }
}
```

Then you can `Core.Thrust.RequestThrottle(val)`.  That gives an API which always keeps the throttle on if the min throttle limit is set, but optionally allows using it to cut the throttle so that zero means zero, or just to bypass the minimums entirely.

In an ideal world, `Core.Thrust.TargetThrottle` would get made private and consumers would switch to using `RequestThrottle`, and would then be explicit about how they expected the throttle to behave under minimum throttle conditions.  Then consumers would either have to use `allowZero` or else call `ThrustOff()` explicitly to shut down the engines.  

BUT what this does is stops the Landing controller from poking around with the MinThrottle and putting all kinds of throttle clamping logic into the Landing controller, which violates separation of concerns.

I don't know what to do about `TransSpdAct` though...

lamont-granquist:
Oh whoops, I made the additional API in #2000 and merged it.

Feel free to make changes to RequestActiveThrottle() if you have to, though, since nothing uses it right now, you won't break anything.

You'll probably want to rebase and force push:

git fetch origin
git rebase origin/dev
git push -f njits23 landing_guidance_bodges

(or however you have or origin + branch setup in your client)

njits23:
Alight, something went wrong somewhere in the implementation of the shutoff after the low deorbit burn, it keeps the engine on, and I can't tell why. That's going to have to be a future me problem to solve.

njits23:
In theory that should've fixed the problem if the no-shutoff after deorbit problem was what I thought it was.

njits23:
Alright. Everything now seems to function as I intended. I would've loved to also be able to get rid of the deorbit burn entirely, but I just couldn't figure that one out.

njits23:
The original fix for the final descent attitude excursions was sloppy. Some craft were still losing control and refusing to fire RCS to correct it, that is now fixed. Aside from that I also cleaned up the logic of the entire section to make more sense, and replaced a TransSpdAct call with a RequestActiveThrottle call.

lamont-granquist:
I'm still not wild with TransSpdAct, but I don't have to time to think of a better API before vacay...

PR #2010, opened 2025-11-05
Fix hotstaging without Δv window open

Nazfib:
Fixes https://discord.com/channels/319857228905447436/331813459417235457/1435745655346954302


lamont-granquist:
Yeah, that makes sense.  Looks like the call is gated behind a pile of booleans that will short-circuit it unless you are actually actively hotstaging as well.

PR #2013, opened 2025-11-21
Make BetterController obey control rate limits

Nazfib:
Before, it only used the max stopping time and minimum flip time when in the PID-controlled range; now, it obeys these limits also in the angular-distance-based control region.


lamont-granquist:
replaced with #2015

PR #2024, opened 2025-12-10
WIP: Fix ComputerModule reinitialization on vessel reload

caseys:
Detects destroyed module references after vessel load (via onVesselGoOffRails), logs in DEBUG, and reruns the MechJeb core initialization so the active vessel rebuilds its modules instead of keeping fake-null MonoBehaviours.

This will help projects like kOS.MechJeb2.Addon acquire valid refs to MechJeb after a save is loaded. 

I am new to MechJeb dev.  Revision advice greatly appreciated.

Claude has assisted in some of this work.

testing now

### DIFF
--- a/MechJeb2/MechJebCore.cs
+++ b/MechJeb2/MechJebCore.cs
@@ -405,11 +405,6 @@ namespace MuMech
 
         public void ReloadAllComputerModules()
         {
-            InitializeComputerModules();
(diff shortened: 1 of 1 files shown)


caseys:
ok!  I figure this out on my end.  MJ doesn't need this change at all.

PR #2025, opened 2025-12-14
Fix min value of velocity in auto turn preventing the rocket from turning (classic ascent)

Clayell:
For some reason, the end velocity cannot be <= 1, or the vessel will not turn. I spent several hours investigating this and did not find out why. It still ends the GRAVITY_TURN mode at the correct speed, but just refuses to turn. Eventually, I just gave up, so the min value is > 1. The new min AutoTurnSpdFactor of 10 (actually 5) leads to a min velocity of about 1.953m/s.

image

Related discussion: https://discord.com/channels/319857228905447436/485125363253641228/1449423156459864286


lamont-granquist:
fixed in #2031

PR #2029, opened 2025-12-16
For hotstaging, sum stages with the same number

Nazfib:
If a stage has ullage motors that light at the same time as the main engine, then hotstaging the _next_ stage does not work properly. Since both the ullage motors and the "regular" engine have the same KSP stage number, hotstaging starts when the ullage motors have less than `HotStagingLeadTime` seconds left to burn, even if the main engine burn time is much longer.

With this change, all burn times with the same KSP stage are added up to calculate the `LastNonZeroDVStageBurnTime`.


lamont-granquist:
yep that makes sense.

PR #2060, opened 2026-01-04
Make PSG Atlas-friendly

AntonKuzin:
Added 2 new fields in the `MechJebModuleStageStats`. The half-stage index is determined automatically, and the end mass value is provided by the user and stored per rocket. Then the new values are passed into the `SimVessel` and used in the `FuelFlowSimulation`.


Nazfib:
It does not work properly for me. If the end mass is set below some threshold, the stage time is completely different from when it is above that value (and the stage's end mass is also wrong). The craft file from these screenshots is attached, with the extension renamed from `.craft` to `.txt`.

Atlas-correct
Atlas-wrong

[Atlas LV3A Agena D.txt](https://github.com/user-attachments/files/25995181/Atlas.LV3A.Agena.D.txt)

PR #2079, opened 2026-02-07
Staging controller improvements

AntonKuzin:
First - just skip any other checks if there already is a reason to decouple.
`InverseStageHasActiveEngines` can sometimes produce the incorrect results, for example, when there's a sustainer stage with boosters - it may decide there are no active engines because the sustainer was activated on the previous stage.


lamont-granquist:
That seems really weird that would be buggy since every R7 or drop-stage atlas out there should have a sustainer stage with boosters?

Do you have a craft file with a replication case?

AntonKuzin:
[Sure](https://github.com/user-attachments/files/25156279/C.zip). It's just a regular souyz without a fairing. Make sure to setup a hotstaging(1s is enough) and use a 4x physical timewarp.
I think the lack of fairing is what exposed this bug - with a fairing the function just returns before reaching this code.

AntonKuzin:
It looks like other people also experience [this bug](https://discord.com/channels/319857228905447436/1115321536807718992/1531294181438394491). Maybe, release the fix now and come up with more elegant solution later?

lamont-granquist:
Rebased (to avoid all the merge commits in the history) and merged in #2295 with one additional `return` fix.

PR #2109, opened 2026-04-07
Fix Linux build, add Nix for environment

markusa380:


lamont-granquist:
one of these days i need to get over the learning curve with nix.

markusa380:
@lamont-granquist I haven't either tbh

PR #2110, opened 2026-04-07
Add larger increments and smooth control to SmartASS

markusa380:
I had both of these features on my wishlist to better control lifting body reentries (Shuttle-style), specifically during the decent control phase using banking.

- Larger increments for attitude adjustments: allow controlling the bank angle without typing in numbers or spamming the - and + buttons. I chose 10deg, but maybe we want to use 5deg? I felt like making it configurable would be a bit overkill
- Add a "smooth control" flag to Surface relative modes, with a degree per second value. If enabled, any changes to the three axes will result in a slow linear transition from the current attitude to the target attitude. I have had too many ships spin out of control because I changed the roll value too quickly while going mach-foo at a 40 degree pitch. I set the default to 10 degrees, which gives a nice smooth transition if enabled.

Now, I would love someone to test this before it gets merged, I did already but don't feel too confident.

### DIFF
--- a/MechJeb2/MechJebModuleSmartASS.cs
+++ b/MechJeb2/MechJebModuleSmartASS.cs
@@ -163,6 +163,8 @@ namespace MuMech
         private Quaternion targetAttitude = Quaternion.identity;
         private AttitudeReference targetReference = AttitudeReference.ORBIT;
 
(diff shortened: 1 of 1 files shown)


lamont-granquist:
Well I didn't get any objections when I asked about this, also no testing, though, but I guess that is what dev is for...

PR #2112, opened 2026-04-08
Fix not being able to launch southwards

Azzurite:
The inclination in the constraint was `Abs()`'d, which prevented the direction vector from being optimized for the negative Z direction. Neither the `Abs()` nor the `ClampPi()` is necessary for the inclination parameter, because the sine functions already do that inherently, so this parameter sanitation was simply removed.

**ALSO**

The `ClampPi` function is actually bugged, as it internally uses `Clamp2Pi`, which clamps from 0 to 6.28. Thus `ClampPI`, even though its description states "Returns the equivalent value in radians between -pi and pi.", is actually only returning values between 0 and 3.14. I'm not sure if I should open this as an issue or anything like that, because I don't really care about it and I haven't noticed any bugs caused by this, but it should probably be changed.

### DIFF
--- a/MechJebLib/PSG/Terminal/FlightPathAngle3Energy.cs
+++ b/MechJebLib/PSG/Terminal/FlightPathAngle3Energy.cs
@@ -21,7 +21,7 @@ namespace MechJebLib.PSG.Terminal
             NumConstraints = 3;
             _gammaT        = gammaT;
             _rT            = rT;
-            _incT          = Abs(ClampPi(incT));
+            _incT          = incT;
         }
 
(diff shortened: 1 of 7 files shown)


lamont-granquist:
Yeah, this winds up going into the initial guess generation via the IncT public accessor on ITerminal.  Think you're wrong about ClampPi() but that is irrelevant and this should work.  I think the guesser is the only consumer of that public accessor as well, so nobody is doing comparisons on IncT where wrapping would matter.

Azzurite:
> Think you're wrong about ClampPi()

Disregard me, I'm dumdum

PR #2113, opened 2026-04-08
BetterController: gain-schedule Kp against maxAlpha

Alex6511:


lamont-granquist:
I think close this one for now given discussion we had on discord

PR #2114, opened 2026-04-08
Delete aircraft autopilot

### LINKED ISSUE #1563: NOTICE:  Aircraft Autopilot is Unmaintaned and Buggy, use the Atmosphere Autopilot mod
The Aircraft Autopilot doesn't need any more bug reports, see:

#1549 
#1494 
#1472 
#1422 
#1420 
#1344 
#1336 
#1335 
#1334 
#1330 
#913 
#625 
#463 

It needs PRs to fix those issues, or else the module should probably be deleted.  Nobody is working those bugs and there is literally years of work left to be done in areas dealing with rockets in MechJeb.

The Atmosphere Autopilot mod is recommended.

NOTE that it'll be necessary to press the P hotkey to disable Atmosphere Autopilot and use MJ (for rockets)

(FAR stability assists should also be turned off for the rest of MJ to work correctly)

Clayell:
Fix https://github.com/MuMech/MechJeb2/issues/1563

It's been years, no one is going to fix aircraft autopilot. Its only use is to confuse players who don't yet know that the wonderful atmosphere autopilot mod exists. I also have no idea what spaceplane autopilot is, as I couldn't seem to access it in mechjeb even without this PR. Regardless, it depends on aircraft autopilot, so I have removed it as well.

The massive changes to csproj are just a git issue, the only real changes are to remove these 4 files from the build. This does compile, but I'm unable to fully confirm that it works as KSP seems to not load the dll for some reason (probably due to me lacking jetbrains annotations)


lamont-granquist:
Yeah, if anyone wants to resurrect and fix it, the code is all still here.

chambm:
Does Atmospheric Autopilot have a runway approach/autoland mode? I haven't been able to find it. Other than that AA is great, but I really need autoland. Just approaching a damn runway without proper HUD guidance is a nightmare.

PR #2116, opened 2026-04-08
Add clickthroughblocker dependency

Clayell:
image
image
works 👍


siimav:
I vote against adding CTB dep. MJ already has a clickthrough blocker implementation which is pretty much equivalent to the one in that mod. One smol issue is fixed here: #2119. The bigger one is map view double clicks deselecting current target but CTB doesn't solve this either. Separate fix for this should go into KSPCF: https://github.com/KSPModdingLibs/KSPCommunityFixes/pull/362

Clayell:
The existing CTB implementation only works for mechjeb's own windows, no? I don't believe it works against windows from other mods

siimav:
Mods do block each others' windows.

lamont-granquist:
Yeah, I got reminded yesterday about how deps and other mods cause headaches.  Particularly because CTB happened to break my install by not loading and causing all the MJ windows to fail.  

Closing this one now in favor of #2119 -- although it isn't a hard 'no', if we find more cases where CTB fixes lots of hard to fix issues then it'd be a good idea to pull it in as a dep rather than DIY.  But if DIY works well enough it is always better than a dep.  More or less this meme:

https://i.programmerhumor.io/2023/02/programmerhumor-io-frontend-memes-javascript-memes-692a23210511647.jpg

PR #2125, opened 2026-04-12
Prevent Reflection errors from other mods tanking MJ

lamont-granquist:
Avoids using the KSP.IO.File.Exist<T>() API that walks the loaded assemblies and throws, looking for the path to the assembly with the type T.

The MuUtils helper replicates the side effect of this API of creating the directory.

Since we construct the path afterwards anyway to load the file the only reason I can see for the reflection-driven-API is for that side-effect, and to cause weird bugs if someone ever moves the DLL location around and makes the two APIs start to disagree.


lamont-granquist:
Successful rage debugging session.

PR #2127, opened 2026-04-14
Actually set rssMode = true when RO is installed

Clayell:
Yet to test in-game, will do so soon (but it does compile). This is probably not the best way to do it but oh well.

### DIFF
--- a/MechJeb2/MechJebModuleAscentSettings.cs
+++ b/MechJeb2/MechJebModuleAscentSettings.cs
@@ -286,7 +286,7 @@ namespace MuMech
         {
             base.OnStart(state);
 
-            if (ForceResetROSettings && ReflectionUtils.IsLoadedRealismOverhaul)
+            if ((int)state > 1 && ForceResetROSettings && ReflectionUtils.IsLoadedRealismOverhaul)
             {
                 ApplyRODefaults();
                 ForceResetROSettings = false;
@@ -323,7 +323,8 @@ namespace MuMech
             DesiredOrbitAltitude.Val = 145000;
             DesiredAttachAlt.Val     = 145000;
 
-            DesiredInclination.Val = Math.Round(Vessel.mainBody.GetLatitude(Vessel.CoMD), 3);
+            if (Vessel != null && Vessel.mainBody != null)
+                DesiredInclination.Val = Math.Round(Vessel.mainBody.GetLatitude(Vessel.CoMD), 3);
 
             Core.Guidance.UllageLeadTime.Val = 20;
 
--- a/MechJeb2/MechJebModuleSettings.cs
+++ b/MechJeb2/MechJebModuleSettings.cs
@@ -42,24 +42,6 @@ namespace MuMech
         [Persistent(pass = (int)Pass.GLOBAL)]
         public bool showAdvancedWindowSettings;
 
-        [Persistent(pass = (int)Pass.GLOBAL)]
-        public bool firstLoad = true;
-
-        public override void OnAwake()
-        {
-            if (!firstLoad)
-            {
-                firstLoad = true;
-
-                bool ROinstalled = AssemblyLoader.loadedAssemblies.Any(a => a.name.Equals("RealismOverhaul", StringComparison.OrdinalIgnoreCase));
-
-                if (ROinstalled)
-                {
-                    rssMode = true;
-                }
-            }
-        }
-
         public override void OnLoad(ConfigNode local, ConfigNode type, ConfigNode global)
         {
             base.OnLoad(local, type, global);
--- a/MechJeb2/MechJebModuleSettings.cs
+++ b/MechJeb2/MechJebModuleSettings.cs
@@ -1,4 +1,6 @@
-﻿using KSP.IO;
+﻿using System;
+using System.Linq;
+using KSP.IO;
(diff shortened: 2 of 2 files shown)


Clayell:
Additionally, in my opinion this should just be relabeled "RO mode". There are many that play RSS without RO and would not benefit from this setting. @lamont-granquist do you have any objections to me changing it to RO mode?

Nazfib:
It should already be defaulting to on if RO is installed:
https://github.com/MuMech/MechJeb2/blob/dev/MechJeb2/MechJebModuleAscentSettings.cs#L330

Clayell:
Just noticed that. I think the problem is that that method continuously spits out a NRE. Looking into it now.
```
[ERR 21:24:49.540] MechJeb module MechJebModuleAscentSettings threw an exception in OnStart: System.NullReferenceException: Object reference not set to an instance of an object
  at MuMech.MechJebModuleAscentSettings.ApplyRODefaults () [0x00195] in <bad6e93192a9446f8215267b3c1c46b5>:0 
  at MuMech.MechJebModuleAscentSettings.OnStart (PartModule+StartState state) [0x00016] in <bad6e93192a9446f8215267b3c1c46b5>:0 
  at MuMech.MechJebCore.OnStart (PartModule+StartState state) [0x000b5] in <bad6e93192a9446f8215267b3c1c46b5>:0 
	UnityEngine.DebugLogHandler:LogFormat(LogType, Object, String, Object[])
	ModuleManager.UnityLogHandle.InterceptLogHandler:LogFormat(LogType, Object, String, Object[])
	UnityEngine.Debug:LogError(Object)
	MuMech.MechJebCore:OnStart(StartState)
	Part:ModulesOnStart()
	<Start>d__323:Part+<Start>d__323.MoveNext_Patch0(<Start>d__323)
	UnityEngine.SetupCoroutine:InvokeMoveNext(IEnumerator, IntPtr)
```

edit:
It's due to Vessel being null in the editor.
image
image

Clayell:
Is there anything that calls [`MechJebModuleAscentMenu.ApplyRODefaults()`](<https://github.com/MuMech/MechJeb2/blob/cf1c18efab483577cdf0614845f80d5660367de5/MechJeb2/MechJebModuleAscentMenu.cs#L61-L67>)? VS2022 shows no references, and commenting it out doesn't interfere with the build (not to be confused with `MechJebModuleAscentSettings.ApplyRODefaults()`, the one with the exception above)

Clayell:
image
Got ApplyRODefaults to not spit an NRE by making it only run in flight. rssMode still isn't being set to true though...

Clayell:
This is quite an annoying problem to solve. I've worked through the unity debugger what the problem is, but I'm not so sure I like the solution.
After going from the editor to launch this is the order of sequences:
[MechJebModuleAscentSettings.OnStart](https://github.com/MuMech/MechJeb2/blob/dev/MechJeb2/MechJebModuleAscentSettings.cs#L285-L294) is called first, which calls ApplyRODefaults()
[MechJebCore.OnLoad](https://github.com/MuMech/MechJeb2/blob/dev/MechJeb2/MechJebCore.cs#L755) runs, which loads the cfg which contains the rssMode = false value
and then rssMode = false

I instead had it set up so that Core.OnSave(null) is called at the end of ApplyRODefaults, but this leads to a different problem
Core.OnSave won't save if `_lastFocus != vessel`, and because we're still kinda in the editor, vessel is still null, so this doesn't actually save to the cfg, leading to the same problem as before

I then set it up so that Core.OnSave didn't need that check, but then I ran into a different problem
When I call Core.OnSave, there are several modules that do stuff based on vessel, so they end up being null, which then causes problems with those other modules

There doesn't seem to be a better method I can start from in MechJebModuleAscentSettings either. OnFixedUpdate just doesn't seem to be called at all, and OnUpdate is only called by MechJebCore after the values are saved for the last time before the launch is loaded, so it's not helpful. OnLoad and OnSave wouldn't be helpful, because the rssMode isn't in MechJebModuleAscentSettings, it's in MechJebModuleSettings.
The only solution I can think of that would work is moving ApplyROSettings() (or just `Core.Settings.rssMode = true;`) into MechJebModuleSettings.OnLoad(), after base.OnLoad is done. This is a bit clunky though.

Clayell:
It seems that the bane of the problems was having `ApplyROSettings()` run when `(int)state > 1`. I added this to get rid of the NRE, but it seems like doing so prevented `ApplyROSettings()` from running until a time where calling `Core.OnSave(null)` would lead to certain other modules being null due to a null vessel. As https://github.com/MuMech/MechJeb2/commit/4f24b68bbd537ff8d3a9ffdf49affe8cd77d67f1 removed the need to have `(int)state > 1`, removing that check allows for `ApplyROSettings()` to run at a time where saving is valid.
Therefore, this PR is no longer needed, but I will note that the behavior in calling ApplyROSettings() is very fragile:
https://github.com/MuMech/MechJeb2/blob/f058d01d4e7d1b7f08556704022fe1210d101927/MechJeb2/MechJebModuleAscentSettings.cs#L285-L294

If there ever comes a time where `ApplyROSettings()` runs when saving is invalid, you will run into the same problem as I had before with other modules being null after loading. @lamont-granquist, perhaps add a note there for the future?

PR #2136, opened 2026-04-24
adding stationary orbit helper to Manuever Planner

b0red-man:
adds `OperationStationary` to `MechJeb2/Manuever`, lets users specify a longitude, then makes nodes to get to a stationary orbit above specific longitude. also allows users to pick longitude using map view, and has basic handling incase stationary orbit cannot be reached.


lamont-granquist:
merged in #2143

PR #2140, opened 2026-05-02
Add experimental standalone Powered Descent Guidance module

gilly262:
## Summary

Draft PR adding an experimental standalone Powered Descent Guidance (PDG) module for airless-body powered descent. 

Intended as a precision/targeted landing tool from a parking orbit and does not replace the existing MechJeb Landing Guidance module. It adds a separate PDG window and backend guidance path using an iterative predictor-corrector powered-descent phase followed by selectable Apollo zero-jerk or gravity-turn terminal guidance. Tests were generally conducted from a parking orbit that overflies the landing site within reason, where it would execute braking and landing burns of 5-10 minutes duration.

Seeking feedback & interest levels in the module. 

## Main changes

- Adds standalone `MechJebModulePoweredDescentGuidance`.
- Adds `MechJebModulePoweredDescentGuidanceWindow`.
- Adds pure solver classes separated from KSP vessel/module state.
- Adds land-at-target and land-anywhere modes.
- Adds basic operational guards and safe-failure status reporting.
- Adds engine and fore-by-throttle RCS thrust accounting.
- Adds PWM throttle mode for engines/configurations that can tolerate rapid on/off throttle commands. (Inspired by SLIM Lander)
- Adds PDG localization keys.

## References

- X. Zhang, D. Wang, T. Wang, and X. Huang, “Dynamic constant thrust general integral guidance for lunar pinpoint landing,” Chinese Journal of Aeronautics, article 104030, 2025. doi:10.1016/j.cja.2025.104030.
- A. R. Klumpp, “Apollo Lunar-Descent Guidance,” MIT/Draper Report R-695, 1971.
- Han, Jo, and Ho 2024, arXiv:2409.01465, for gravity-turn terminal guidance.

## Testing

Tested the standalone Powered Descent Guidance module in stock KSP 1.12.5 and RSS/RO/RP-1 Sol sandbox.

| ID | Environment | Body | Mode | Terminal | Result |
|---|---|---|---|---|---|
| 1 | Stock | Mun | Land anywhere | Gravity Turn | Pass |
| 2 | Stock | Mun | Targeted | Apollo | Pass |
| 3 | Stock | Kerbin | Atmospheric guard | n/a | Pass |
| 4 | Stock | Val | Land anywhere / poor geometry | Gravity Turn | Safe failure |
| 5 | Stock | Val | Land anywhere / feasible orbit | Gravity Turn | Pass |
| 6 | RSS/RO | Moon | Targeted LDME | Apollo | Partial / min-throttle limitation |
| 7 | RSS/RO | Moon | Land anywhere inclined orbit | Gravity Turn | Pass |
| 8 | RSS/RO | Moon | RCS fore-by-throttle only, Land anywhere | Apollo | Partial / RCS terminal issue |
| 9 | RSS/RO | Moon | PWM Land Anywhere | Gravity Turn| Pass |

## Known limitations

- Experimental, airless-body-only guidance.
- No atmospheric drag, lift, parachute, or aerobraking modelling.
- Minimum-throttle / over-thrust cases are not fully solved. Gravity-turn terminal guidance has behaved better than Apollo terminal guidance in some of these cases.
- RCS fore-by-throttle support is approximate and needs more testing with less deliberately configured RCS layouts.
- Some infeasible geometries are expected to fail safely rather than recover.
- PWM control is rudimentary and subject to KSP fixed-timestep limitations. Tends to behave acceptably in testing.


lamont-granquist:
I've been working for the past week or two on redoing the suicide burn countdown timer which grew its own autopilot and now pretty much does the same thing:

#2141

gilly262:
Thanks, I had a look.

There could be overlap for the terminal phase, but I think the scopes are different.

This is aimed at targeted landing from orbit, including predictive coast, powered descent ignition and subsequent targeting (ie: braking from a parking orbit, controlling throttle and steering) in both descent and terminal phases. In testing this was from a downrange in the order of ~400-600km in a RSS Sol Apollo LM-style craft case.

lamont-granquist:
Yeah, I think I can't accept this.  I can't handle maintaining two different competing versions of landing guidance, and the problem with the choices here are that they're sort of evolutionary dead ends.  It doesn't help me either fix or clean up the existing landing guidance, it just gives me even more code to deal with.  I'm also familiar with the Han, et al reference and that is actually the code behind the analytic suicide burn countdown that I'm removing.  That's a flat earth, constant gravity model and is kind of dead end.  It'll have sharp edges that'll never get any better in the general case.  The APDG itself is also a dead end.  It'll never support and aero model, and it'll fail for arbitrary kinds of spacecraft that people throw at it, which I'll wind up being the one who needs to deal with.

There's also other problems with this contribution in that it is correctly split up to separate UI from control from pure math, but the pure math needs to go into MechJebLib, which is a non-Unity library with its own vector classes, etc and it has unit tests in MechJebLibTests, plus it is duplicating function like IsFinite and printing vectors which already have solutions in the codebase.

So, I can't accept this.  I've considered APDG before, but it doesn't get me anywhere.  I'm literally ripping the analytical GT formulas out of the codebase to replace them with a higher fidelity numerical solution with central gravity, etc.

So unfortunately, I'm going to close this.  In general, it would have been better to open an issue and ask up front about this kind of effort before engaging in it.

I know that the existing landing guidance here kind of sucks, but the problem that I want solved is to address that, and either fully solve that issue with a clean rewrite (with aero and parachutes and solving problems like multi-stage landings and have a path to F9-style boostback burns in RSS/RO, which is what everyone really wants), or else to do the hard work of refactoring the existing landing guidance into the shape where its parts are fully disentangled and it can actually be incrementally improved.

gilly262:
Understood, this was a learning project for me, so thought I'd throw it out there. Thanks for the insights with where you're at with the landing problem and the points on the code implementation (I completely overlooked MechjebLib, my bad), it's something I don't have much exposure on yet with the consequences to maintenance, generality and direction/evolution. I'll delete the branch, and consider contributing perhaps with some smaller issues in the future.
Cheers!

MechJeb2/PoweredDescentGuidance/PDGTargeting.cs
```diff
@@ -0,0 +1,229 @@
...
+        private static double GetTerrainAltitude(CelestialBody body, double lat, double lon)
+        {
+            if (body.pqsController != null)
+            {
+                Vector3d radial = body.GetRelSurfaceNVector(lat, lon);
+                return body.pqsController.GetSurfaceHeight(radial) - body.pqsController.radius;
+            }
+            return body.TerrainAltitude(lat, lon);
+        }
+    }
```
lamont-granquist, line 226: You might want to look at CelestialBodyExtensions and TerrainAltitude() and GetPQSSlopeDegrees() and quite possibly your implementation here is better.  I'm thinking my GetPQSSlopeDegrees() is probably named wrong as well, since that seems to not use PQS to get the slope, instead of using PQS to get the slope.

I'm kind of thinking that the extension methods in MJ could use helpers to take lat and lng or alternatively a world vector and get the altitude back, and optionally the slope--and then do the right thing to get it from pqs if you have it or not if you don't.  That could avoid the finite differencing in the current slope calculation that i just added.

That would be a good PR, I'm looking at this and kinda want it.  I don't know the exact shape the API should be but getting the slope is less common, possibly just a ref parameter that could be ignored (but then if you ignore it you might still wind up doing finite differencing even though you don't need it).

sarbian:
@gilly262 MJ also allow a module to be in a separate DLL. You could publish your landing code as a plugin for MJ. This would make it available now for those who want an improved landing module now and can deal with the limitation of the method.

gilly262:
@sarbian I had considered this and might do so, thanks for the thought.

Issue #1910, opened 2024-10-09
Feature request: deleting all maneuver nodes should change Smart ASS from NODE to KILL ROT


jessicest:
Current behavior: when all nodes are deleted, Smart ASS will stay in NODE mode but it won't do anything
Desired behavior: it should swap to KILL ROT mode

Rationale: if the user requested a Smart ASS mode, they're more likely to want Smart ASS to keep doing *something* rather than to suddenly give up and do nothing.

Specific use-case: when manually executing a Principia node with Smart ASS, Principia likes to delete the maneuver node as soon as the burn "should" be completed. In this scenario, Smart ASS would be more helpful if it locked rotation than if it quietly turned itself off.

PR #2146, opened 2026-05-03
add smartASS killrot fallback

b0red-man:
addresses #1910, sets smartASS to killrot if all nodes were deleted and was in NODE mode

### DIFF
--- a/MechJeb2/MechJebModuleSmartASS.cs
+++ b/MechJeb2/MechJebModuleSmartASS.cs
@@ -164,7 +164,7 @@ namespace MuMech
         private AttitudeReference targetReference = AttitudeReference.ORBIT;
 
         private static readonly double LARGE_INCREMENT = 10.0;
-
+        private int lastNodeCount = 0;
         // Current attitude/direction (for smooth interpolation)
         private Vector3d curDirection = Vector3d.zero;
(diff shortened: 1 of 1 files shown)


MechJeb2/MechJebModuleSmartASS.cs
```diff
@@ -224,6 +224,24 @@ public override void OnLoad(ConfigNode local, ConfigNode type, ConfigNode global
                 Engage(false);
         }
 
+        public override void Drive(FlightCtrlState s)
```
lamont-granquist, line 227: Does this need to be in Drive() as opposed to OnFixedUpdate() later in the code?

If it does, it is probably worth a short comment on why.  It is mostly just touching internal state and not the flight controls (which is what is in my mental model of why you need to use Drive(), but the details on that kinda paged out of my brain a few years ago).

b0red-man: I put this in `Drive()` so the check only runs when SmartASS is active. Switching to `OnFixedUpdate()` would require more checks to make sure the module is active (and also to prevent edge cases where it switched to KILLROT even when the module isn't active)
I can change the code to use OnFixedUpdate if you prefer.

PR #2148, opened 2026-05-04
Eliminate UI ExpandWidth allocs

siimav:
There are already helpers in place but half of the code isn't using those.


siimav:
Hmm that's an odd diff
image

lamont-granquist:
Yeah, I have no idea why those ZWNBSP things started showing up.  Probably some weird setting in some Rider update or I just configured something wrong, IDK.  I resync'd a bunch of .editorconfig stuff in a stealth merge in the hoverslam stuff and tweaked a few rules that were starting to get really annoying.  The way that .editorconfig and Rider/Resharper rules have seemed to be evolving have been pretty annoying.

PR #2186, opened 2026-05-14
Fix launch to plane with negative inclination

### LINKED ISSUE #2184: Ascent guidance holds on to negative inclination with `launch into plane` enabled
## MechJeb Version
2.15.1.0-1584

## KSP Version
1.12.5.3190 (running on Proton 11)

## Description
If I launch to target plane, a burn profile that would otherwise be normal winds up looking like the first screenshot of night side earth. 

Process, using PSG:
1. set up parameters (also attached)
2. select target (moon)
3. `launch into plane of target`
4. `engage autopilot`

Doing the same setup will produce the same results on revert. I've tried it with more than 1 fixed stage, same result. 

The only ways I have found to avoid this are
- abort launch into plane, disengage, and reengage. 
- make `Orbit Inc.` positive before or during the warp.

The second screenshot of dayside earth is without launch into plane selected, and shows the expected result

## Replication Case

Image

Image

Image

Nazfib:
Since #2112, the inclination fields in the constraints include the sign, which is used to signal to the initial guesser that the initial launch azimuth should be southwards. When passing this signed inclination to the functions for calculating the angular momentum and eccentricity vectors, this results in those vectors being calculated for an orbit with a LAN 180° offset from the specified orbit.

This PR makes sure that the absolute value of the inclination is taken, anywhere it is used for astrodynamics calculations; the initial guesser is the only place where the inclination with sign is used. Fixes #2184.

### DIFF
--- a/MechJebLib/PSG/Terminal/FlightPathAngle3Energy.cs
+++ b/MechJebLib/PSG/Terminal/FlightPathAngle3Energy.cs
@@ -33,7 +33,7 @@ namespace MechJebLib.PSG.Terminal
         {
             double gammaT = _gammaT;
             double rT     = _rT;
-            double incT   = _incT;
+            double incT   = Abs(_incT);
 
             var rf = V3.CopyFromIndices(x, ri);
             var vf = V3.CopyFromIndices(x, vi);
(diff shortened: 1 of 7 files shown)


lamont-granquist:
That almost makes me think the right fix would be to push those into the Astro functions so that they all support the MJ negative inclination convention without flipping the LAN around...

PR #2201, opened 2026-05-20
Github action build workflow

sarbian:
First take for the Action workflow. Simple one for now to see how it goes. 

Tests are disabled since they where failing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,78 @@
+name: build
+
+on:
+  push:
+    branches: [master, dev]
+    paths:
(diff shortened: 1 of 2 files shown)


lamont-granquist:
Yeah, I've been doing a bunch of work to fix my failing random tests and such to make them more reliable (tends to throw a lot of spurious random errors right now).

I think there are a few failures, though, which require running the mono vm that is inside KSP instead of the .net runtime, which are sort of actual failures highlighting differences between the runtimes.  The Falcon9GTO example reliably fails on a different runtime.

Issue #2118, opened 2026-04-10
Ascent Guidance to Plane of Target Error


ZeevPack:
## MechJeb Version
2.15.2.0

## KSP Version
1.12.5.3190

## Description
When using Ascent Guidance and using the function to launch into the plane of target in the same SOI, time warp stops at incorrect time, and automatic launch occurs far from being in plane with target causing inaccurate final orbit compared to target.

## Replication Case
Separate 2 docked vessels in low circular Mun orbit, ~30km altitude. Land using one at a lesser inclination than the still-orbiting-vessel. Target orbiting vessel. Use Ascent Guidance's function to launch into plane of target for 10 km altitude. Engage autopilot with Auto-Warp activated. Launch will occur at a substantially inaccurate time before launch site is under target's orbit.

## KSP.log
https://www.dropbox.com/scl/fi/n327p30kcmpeorb9j0f9y/Player.log?rlkey=ux57qu0uokkdgaff8cwyilfj2&st=izb1rzlp&dl=0

NOTE: The latest development release of MechJeb can be downloaded from jenkins ( https://tinyurl.com/hgckqmk ).

lamont-granquist:
If you have "Override Warp to Plane" checked that is the intended function of that button.

ZeevPack:
I did not have that checked. Please see the attached video of me duplicating the error. The target became unselected after ascent began, but the fact it warped and initiated to the wrong time to be able to launch into the correct plane is what I think is at issue.

https://www.dropbox.com/scl/fi/4dl6svqwxr7oob6h1co9c/2026-04-11-15-26-01.mp4?rlkey=pkuoxjma6smsg5pew69ggux83&st=qx1zzk0v&dl=0

ticktricktrack:
I have the same issue trying to rendezvous with a satellite from Kerbin. When I warp manually first then engage autopilot it works. But autowarp goes to a completely wrong time.

Nazfib:
> I did not have that checked. Please see the attached video of me duplicating the error. The target became unselected after ascent began, but the fact it warped and initiated to the wrong time to be able to launch into the correct plane is what I think is at issue.
> 
> https://www.dropbox.com/scl/fi/4dl6svqwxr7oob6h1co9c/2026-04-11-15-26-01.mp4?rlkey=pkuoxjma6smsg5pew69ggux83&st=qx1zzk0v&dl=0

[...]

Issue #985, opened 2018-02-07
Ascent Guidance miscalculates rendezvous time


EoD:
When I press "Launch to Rendezvous", the launch is too late and the orbiting partner is long gone when I reach it. It does improve, when I increase "Orbit altitude" to a higher value.
![image](https://user-images.githubusercontent.com/293499/35943203-9570fdc6-0c58-11e8-8912-6c7143dc4953.png)
![image](https://user-images.githubusercontent.com/293499/35943346-f852ee90-0c58-11e8-862c-a40ecea59004.png)

I am using Mechjeb 2.7.0.

As there seem to be a (missing?) correlation between the orbit altitude and the rendezvous timer, would it be possible to auto-set the orbit altitude to something similar to the "target altitude/apoapsis"? This would avoid unnecessary fuel consumption.

sarbian:
This function needs the vessel to be launched once untill circularisation for MJ to know the time to reach orbit . Then you do do a revert to launchpad and use the launch to rdv.

EoD:
@sarbian I am a bit confused. I did launch the vessel and quickloaded (as Minmus has no spaceports), but it didn't work. What exactly do you mean?

sarbian:
Hum, I am not sure it would work when launching from anywhere. The recoreder may not start... 
I'll have to check

EoD:
Can you explain to me (I have no clue), why it is not possible to calculate an estimated rendezvous by using the targets orbit (median of periapsis/apoapsis), the current position (or just the angle?) and the available thrust?
Of course assuming that there is no atmosphere.

sarbian:
Where did I say it is not possible ? The code is not there. If someone want to write it they are welcome to do it. 
Currently we have a solution that work well for most needs (launch from a space center) and the rendez-vous autopilot for other cases.

dlrk2:
Does the launch and revert have to be done every launch, or can an old/original number be reused for a similar launch vehicle?

sarbian:
It saves the value for a vessel name.

lamont-granquist:
It is entirely possible but the problem of doing it without a prior rocket launch is pretty much a post-grad level trajectory optimization problem -- particularly the "without overshooting" bit.  I've spent the past 6-8 months off and on trying to teach myself how to do it, and I took graduate level quantum mechanics in college...

lamont-granquist:
#1061 is the rewrite that will eventually be able to solve this right, closing as a dup of that.

PR #2206, opened 2026-05-25
Remove Launch to Rendezvous

lamont-granquist:
This only has any hope of working in pure equatorial-to-equatorial and fails for anything else.

While it does actually work fine for the first launches that typical players do to go to the Mun or rendezvous with an equatorial space station, I think it is teaching people a fundamental dead end.

Users should be launching into a phasing orbit in the correct plane, and then doing a rendezvous.

At some point PSG should support direct ascents to rendezvous with the Moon/Mun or a space station.  In the general case this is a legitimately difficult optimization problem to set up.

### DIFF
--- a/MechJeb2/CachedLocalizer.cs
+++ b/MechJeb2/CachedLocalizer.cs
@@ -22,7 +22,6 @@ namespace MuMech
             MechJebAscentButton11,
             MechJebAscentButton12,
             MechJebAscentButton13,
-            MechJebAscentButton14,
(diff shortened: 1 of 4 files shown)


lamont-granquist:
See #2118 for launch-to-rendezvous and the magic non-zero launchPhaseAngle showing up and trolling people.

lamont-granquist:
See also like #985 for another example of how users don't understand how to use this feature.

PR #2221, opened 2026-06-01
Izzo solver rewrite / test changes

lamont-granquist:
Replace Izzo solver with one originating from the poliastro (MIT License) implementation.


lamont-granquist:
This started out as the poliastro solver, but I threw away the mMax, MinTime stuff to instead just let the Householder iteration fail if the problem isn't feasible.  I also lifted (only) the initial guess equations from pykep.

The sense of the lowpath boolean has been changed to remove a non-differentiable condition in the equations.

This still uses prograde/retrograde convention around the z-axis, that needs to be changed at some point to provide a more unified API with the Gooding solver (prograde defined by h-vec of initial orbit), and I need shortway/longway without any non-differentiable branches between polar orbits, so this implementation isn't done yet.

lamont-granquist:
Tests were also fixed to be pseudo-random instead of random and to include tests of multi-rev cases where the random orbits were elliptical.

PR #2240, opened 2026-06-15
Remove VesselState MovingAverage values

lamont-granquist:
I suspect that these are MovingAverages either due to very old 2012-era KSP having more single precision floats than doubles and this was a workaround to precision jittering, or else it was just premature optimization.

There's other places where the moving averages are deliberately being used to smooth out behavior.

I might be wrong and that the low pass input filter is necessary for some control theory reasons, but I'd rather find that by breaking it and push that requirement for smoothing explicitly down into the consumers.

I suspect that this might actually fix some bugs.


lamont-granquist:
I think they were introduced back in f85d41fc48e39659fec17aaed0ce071a05fd29e3

PR #2269, opened 2026-06-24
Convert PSG mass constraints to hermite-simpson

lamont-granquist:
The old constraints don't take into account throttle, so anything less than maxthrottle was always using up just as much mass for less thrust applied in vdot, which is always a bad trade.

### DIFF
--- a/MechJebLib/Utils/AutoDiff.cs
+++ b/MechJebLib/Utils/AutoDiff.cs
@@ -413,6 +413,128 @@ namespace MechJebLib.Utils
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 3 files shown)


MechJebLib/Utils/AutoDiff.cs
```diff
@@ -797,10 +919,14 @@ public static int ApplyHermiteSimpsonDynamics(double[] f, alglib.sparsematrix j,
...
                     if (lastindex == index)
                         continue;
-                    alglib.sparseappendelement(j, indexes.Index(k), jacZ[k]);
+                    alglib.sparseappendelement(j, index, jacZ[k]);
                     lastindex = index;
                 }
 
+            jacX.Dispose();
+            jacY.Dispose();
+            jacZ.Dispose();
```
lamont-granquist, line 928: this also cleaned up ~100MB of garbage per run, still at around 600MB of garbage per run though.

PR #2290, opened 2026-07-14
Update zh-cn localization: add missing keys and fix untranslated entries

xiaolingtong:
Update of zh-cn Localization
- Added 5 missing translation keys (TargetTimeToAN, TargetTimeToDN, TargetMeanAnomaly, TargetTrueLongitude, TargetLDN)
- Translated 33 previously untranslated entries
- Fixed 1 mistranslation (SmartASS_button36)
- Fixed key name mismatch (Scrmsg → Scrmsg1)
- Translated 11 additional partially/fully untranslated entries

### DIFF
--- a/Localization/zh-cn.cfg
+++ b/Localization/zh-cn.cfg
@@ -2,41 +2,41 @@ Localization
 {
     zh-cn
     {
-        #MechJeb_ModuleAreLocked = 部分模块被禁用，直到你在科技树中解锁相应节点或升级追踪站。
+        #MechJeb_ModuleAreLocked = 部分模块功能关闭直到你解锁科技树的关键性节点和升级追踪站.
         #MechJeb_OnlineManualbutton = 在线手册
 
-        #MechJeb_InstallCheckA_title = MechJeb2 安装不正确
-        #MechJeb_InstallCheckA_msg = MechJeb2 安装不正确，无法正常工作。\n所有 MechJeb2 文件应位于 KSP 中如下目录 \n<KSP>\n\tGameData\n\t\tMechJeb2\n\t\t\tParts\n\t\t\tPlugins\n\n请勿移动 MechJeb2 文件夹中的任何文件。\n\n不正确的路径：\n
+        #MechJeb_InstallCheckA_title = 错误的Mechjeb2安装方式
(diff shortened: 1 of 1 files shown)


Localization/zh-cn.cfg
```diff
@@ -46,211 +46,309 @@ Localization
...
+        #MechJeb_Hohm_title = 双脉冲 (霍曼)转移到目标
+        #MechJeb_Hohm_intercept_only = 仅拦截, 不计算捕获 (撞击/飞掠)
+        #MechJeb_Hohm_simpleTransfer = 简单共面霍曼转移
+        #MechJeb_Hohm_Label1 = 微调目标周期偏移
+
+        #MechJeb_Hohm_Exception1 = 必须为霍曼转移选择目标。
+        #MechJeb_Hohm_Exception2 = 进行霍曼转移的目标必须在同一引力范围内
+        #MechJeb_Hohm_Exception3 = 与目标的升交点不存在.
+        #MechJeb_Hohm_Exception4 = 与目标的降交点不存在.
+        #MechJeb_Hohm_Exception5 = 与目标的降交点与升交点都不存在.
```
lamont-granquist, line 131: Was this section based on translating the old Localizations, and not the ones updated by https://github.com/MuMech/MechJeb2/pull/2087/changes ?

xiaolingtong: Hi, sorry I didn't notice earlier. I've now realigned the Chinese localization against the latest en-us localization file in the repo, and added the missing modules (Hoverslam, stationary orbit, transfer to another planet, etc.).

Localization/zh-cn.cfg
```diff
@@ -227,30 +232,30 @@ Localization
     //Ascent Guidance
         #MechJeb_Ascent_title = 自动发射
         #MechJeb_Ascent_ascentPathList1 = 经典的上升配置
-        #MechJeb_Ascent_ascentPathList3 = 序列二次规划法（SQP）引导（RSS/RO）
+        #MechJeb_Ascent_ascentPathList3 = 初矢量制导 (RSS/RO)
```
lamont-granquist, line 235: I think that translates roughly as "primer vector guidance" but in dev that has been replaced by "Polynomial SQP guidance" so there would need to be a different translation there.

(Also, as an aside, I think "primer vector guidance" would be "主矢量制导" and you can find papers on it by googling "主矢量" "lawden" -- while I get nothing from "初矢量" "lawden").

lamont-granquist:
Well this is probably good enough to merge, we can fix the PSG label later.

PR #2302, opened 2026-08-13
Fuel flow simulation and staging controller improvements

AntonKuzin:
See the comments for details

### DIFF
--- a/MechJeb2/MechJebModuleStagingController.cs
+++ b/MechJeb2/MechJebModuleStagingController.cs
@@ -613,7 +613,7 @@ namespace MuMech
                         PartResource r = p.Resources.Get(propellant.id);
 
                         if (r.amount <= p.resourceRequestRemainingThreshold)
-                            continue;
+                            return false;
                         if (r.info.id == PartResourceLibrary.ElectricityHashcode)
                             continue;
                         if (!tankResources.Contains(r.info.id))
--- a/MechJebLib/FuelFlowSimulation/FuelFlowSimulation.cs
+++ b/MechJebLib/FuelFlowSimulation/FuelFlowSimulation.cs
@@ -129,7 +129,7 @@ namespace MechJebLib.FuelFlowSimulation
             ComputeRcsUllageTime(vessel);
 
             UpdateResourceDrainsAndResiduals(vessel);
-            double currentThrust = vessel.ThrustMagnitude;
+            int activeAngines = vessel.ActiveEngines.Count;
 
             for (int steps = MAXSTEPS; steps > 0; steps--)
             {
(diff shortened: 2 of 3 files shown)


MechJeb2/MechJebModuleStagingController.cs
```diff
@@ -613,7 +613,7 @@ private bool HasActiveOrIdleEngineOrTankDescendant(Part p, List<int> tankResourc
                         PartResource r = p.Resources.Get(propellant.id);
 
                         if (r.amount <= p.resourceRequestRemainingThreshold)
-                            continue;
+                            return false;
```
AntonKuzin, line 616: This is for crossfeeding boosters. Currently, it doesn't stage till both fuel **and** oxidizer are depleted. But it's not optimal - it should stage if **one** of them is empty

MechJebLib/FuelFlowSimulation/FuelFlowSimulation.cs
```diff
@@ -388,7 +388,7 @@ private double MinimumTimeStep()
         {
             double maxTime = ResourceMaxTime();
 
-            return maxTime < double.MaxValue && maxTime >= 0 ? maxTime : 0;
+            return maxTime < double.MaxValue && maxTime > 0.001 ? maxTime : 0.001;
```
AntonKuzin, line 389: This should address the steps limiter issue. For cryogenic fuels the amount of fuel left per tank is slightly different, so it tries to calculate the fuel flow in 1e-6 steps.

MechJebLib/FuelFlowSimulation/FuelFlowSimulation.cs
```diff
@@ -140,13 +140,12 @@ private void SimulateStage(SimVessel vessel)
 
                 // FIXME: if we have constructed a segment which is > 0 dV, but less than 0.02s, and there's a
                 // prior > 0dV segment in the same kspStage we should add those together to reduce clutter.
-                if (Abs(vessel.ThrustMagnitude - currentThrust) > 1e-12)
+                if (dt >= 0.02 && activeAngines != vessel.ActiveEngines.Count)
```
AntonKuzin, line 141: This should address the phantom 0 dv 0 burn time phases constantly appearing and disappearing.

MechJebLib/FuelFlowSimulation/SimResource.cs
```diff
@@ -27,8 +27,8 @@ public double Amount
         public SimResource Drain(double resourceDrain)
         {
             _amount -= resourceDrain;
-            if (_amount < 0)
-                _amount = 0;
+            if (_amount < ResidualThreshold)
+                _amount = ResidualThreshold;
```
AntonKuzin, line 31: As we may drain more fuel than we should, just don't allow it.

PR #2305, opened 2026-09-13
Restore engine thrust limits after differential throttle

maddavo:
Differential throttle writes each `ModuleEngines.thrustPercentage` while solving for torque. Its previous cleanup ran only when the GUI toggle changed from enabled to disabled, and reset every engine to 100% rather than restoring the value that existed before differential throttle was used.

This change records an engine's limiter before the first solver write and restores that value whenever Differential Throttle becomes inactive, a vessel is docked or undocked, or the controller is destroyed. It therefore prevents the controller from leaving asymmetric solver output as a vessel's persistent thrust limit, while preserving intentional non-100% engine limits.

A vessel already saved with stranded limits still needs those limits restored once; this does not try to guess whether an existing saved limiter was intentional.

Validation: `dotnet build MechJeb2.sln -c Release --no-restore` succeeded in the configured KSP checkout.

### DIFF
--- a/MechJeb2/MechJebModuleThrustController.cs
+++ b/MechJeb2/MechJebModuleThrustController.cs
@@ -163,6 +163,12 @@ namespace MuMech
 
         public DifferentialThrottleStatus DifferentialThrottleSuccess = DifferentialThrottleStatus.SUCCESS;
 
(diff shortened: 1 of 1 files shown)


lamont-granquist:
this LGTM

